PR #194, opened 2015-02-10
version bumps

SethTisue:
[Review on Reviewable](https://reviewable.io/reviews/netlogo/galapagos/194)

### DIFF
--- a/app/controllers/CompilerService.scala
+++ b/app/controllers/CompilerService.scala
@@ -15,9 +15,8 @@ import
   com.fasterxml.jackson.core.JsonProcessingException
 
 import
-  org.nlogo.{ api, core, tortoise },
-    api.CompilerException,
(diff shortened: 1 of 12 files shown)


SethTisue:
hmm. I don't understand why the tortoise artifact isn't resolving.

TheBizzle:
I'll take care of fixing the resolution issue.

TheBizzle:
As far as I can tell, the problem is sbt/sbt#1820.

TheBizzle:
For the time being, I've switched to some BinTray repositories that don't mirror to JCenter (to avoid having Typesafe Releases block us from getting our artifacts).  With that, everything works fine again.  Hopefully one day we can go back to putting our artifacts on JCenter.  (But that day is not today.)

PR #200, opened 2015-02-26
Close #179: fix link thickness rendering

### LINKED ISSUE #179: Link thickness drawing incorrectly; it only changes the width of the link
Here is "Info Tab Example" in JVM NetLogo:

jvminfotab

Here it is in Tortoise, where the bottom-most turtle protrudes from the bottom of the arrow:

tortoiseinfotab

The labels prove that the turtles are in the correct locations.  The problem appears to be that link thickness in Beak is only changing the **width** of the drawn link, whereas it should also be extending the link's display towards the link ends.

skorlir:
Some caveats apply.

Covers most common cases where link extension occurs; doesn't cover cases such as, for instance, when the link wraps but the turtle at the end doesn't. Nor does it cover @qiemem 's weird case where the link wraps in all 3 directions but the turtle stays in the corner of the canvas.

[Review on Reviewable](https://reviewable.io/reviews/netlogo/galapagos/200)

### DIFF
--- a/app/assets/javascripts/TortoiseJS/agent/view.coffee
+++ b/app/assets/javascripts/TortoiseJS/agent/view.coffee
@@ -257,6 +257,28 @@ class TurtleView extends View
       x2 = end2.xcor
       y2 = end2.ycor
 
+      if link.thickness > @onePixel
+        shortestX = Math.min(x1 - x2, x2 - x1)
+        shortestY = Math.min(y1 - y2, y2 - y1)
+
+        theta = Math.atan2(shortestY, shortestX)
+        xd    = Math.abs(link.thickness * Math.cos(theta)) / 2
+        yd    = Math.abs(link.thickness * Math.sin(theta)) / 2
+
+        if 0 < link.heading < 180
+          x1 -= xd
+          x2 += xd
+        else
+          x1 += xd
+          x2 -= xd
+
+        if 90 < link.heading < 270
+          y1 += yd
+          y2 -= yd
+        else
+          y1 -= yd
+          y2 += yd
+
       wrapX = canWrapX and ((x1 - (x2 - @patchWidth) < Math.abs(x1 - x2)) or (x2 - (x1 - @patchWidth)) < Math.abs(x1 - x2))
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/TortoiseJS/agent/view.coffee
```diff
@@ -248,17 +248,37 @@ class TurtleView extends View
     @ctx.lineTo(x2,y2)
 
   drawLink: (link, turtles, canWrapX, canWrapY) ->
+    shouldWrapInDim = (canWrap, worldMin, worldMax, cor1, cor2) ->
+      dist    = Math.abs(cor1 - cor2)
+      dimSize = worldMax - worldMin + 1
```
qiemem: Dimensions are stored as `@patchWidth` and `@patchHeight`. Should probably use those instead (so, take an argument `dimension` or something)...

app/assets/javascripts/TortoiseJS/agent/view.coffee
```diff
@@ -248,17 +248,37 @@ class TurtleView extends View
...
+      canWrap and dist > dimSize / 2
```
qiemem: Way better than the previous inequality. Thanks!

app/assets/javascripts/TortoiseJS/agent/view.coffee
```diff
@@ -248,17 +248,37 @@ class TurtleView extends View
...
+
+    adjustLinkEnds = (end1Cor, end2Cor, trigShift, minForExtending, maxForExtending) ->
+      diff = Math.abs(link.thickness * trigShift) / 2
```
qiemem: Could probably be done with `atan2`, but whatevs.

skorlir:
:tada: This is my first merged pull request ever with the CCL! (That took a long time, considering I never technically left from the Summer.)

TheBizzle:
:tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada:
:tada: Party hats and confetti for everyone!!!  :tada:
:tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada::tada: :tada: :tada: :tada:

PR #202, opened 2015-03-01
Add command center with output

qiemem:
[Review on Reviewable](https://reviewable.io/reviews/netlogo/galapagos/202)


TheBizzle:
Looks pretty good.  How about squashing it down so that the mouse stuff is in one commit and everything else is in another?

qiemem:
Done!

I also notice that github now properly replaces the list of commits after a force push, so that's cool.

PR #205, opened 2015-03-05
CompilerService loads models from filesystem instead of via HTTP

mrerrormessage:
I made this fix yesterday when attempting to use JMeter. JMeter tests run against the `/compile-url` path with the model url referring to something on the server resulted in JMeter crashing the server. This fixes that crash and should generally improve the load time of `/compile-url`.

[Review on Reviewable](https://reviewable.io/reviews/netlogo/galapagos/205)

### DIFF
--- a/app/controllers/CompilerService.scala
+++ b/app/controllers/CompilerService.scala
@@ -4,7 +4,7 @@ import
   java.net.{ MalformedURLException, URL }
 
 import
-  scala.util.Try
+  scala.util.{ Try, matching }, matching.Regex
 
 import
   scalaz.{ NonEmptyList, Scalaz, Validation, ValidationNel },
@@ -21,12 +21,15 @@ import
       CompiledModel.CompileResult
 
 import
-  play.api.{ cache, libs, mvc, Play },
-    cache.Cache,
-    libs.json.{ Json, JsObject, Writes },
-      Json.toJsFieldJsValueWrapper,
-    Play.current,
-    mvc.{ Action, Controller, RequestHeader, Result }
+  play.{utils, api},
+    utils.UriEncoding,
+    api.{ cache, libs, mvc, Play },
+      cache.Cache,
+      libs.json.{ Json, JsObject, Writes },
+        Json.toJsFieldJsValueWrapper,
+      Play.current,
+      mvc.{ Action, Controller, RequestHeader, Result }
+
 
 import
   controllers.PlayUtil.EnhancedRequest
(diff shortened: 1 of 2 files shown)


TheBizzle:
This strikes me as a less-than-ideal solution to the problem.  Have you tried reverse routing to `controllers.Assets`?  I bet it does these kind of lookups really fast, since assets are loaded onto the classpath the `Assets` router even caches them.

mrerrormessage:
I'll give that a try.

app/controllers/CompilerService.scala
```diff
@@ -203,14 +205,23 @@ object CompilerService extends Controller {
...
-      case _: MalformedURLException => s"'$url' is an invalid URL.".failureNel
-    } getOrElse {
-      s"An unknown error occurred while processing the URL '$url'. Make sure the url is publicly accessible.".failureNel
-    }
+  private def fetchURL(url: String, hostUri: String): ValidationNel[String, String] = {
+    val locallyHostedRegex = s"^https?://$hostUri([A-Za-z0-9%/]+\\.nlogo)$$".r
+    locallyHostedRegex.findFirstIn(url).flatMap {
+      case locallyHostedRegex(file) =>
+        Assets.resourceNameAt("public", file).map(path =>
+          usingSource(_.fromFile(path))(_.mkString).successNel
```
TheBizzle: Bah, using `Assets` still leaves you reading a file from the disk unless you want to use `Assets.at` and bubble an `Action` up the call stack (which you probably don't want to do).

The point of my earlier suggestion was to leverage the fact that Play already makes asset loading fast by putting the assets on the classpath.  I guess I would suggest grabbing the file from the classpath directly with `Play.resource(s"/public/$file")` (which returns you its classpath URL, which I would expect you to be able to read from speedily with `usingSource`).

mrerrormessage: Yeah. I was continuing to work on that locally, but the options seemed to be using package-private methods (which is gross) or bubbling an Action up the call stack (which seemed even worse). I'll give it another try using `Play.resource`, which I tried briefly, but didn't seem to work as I had expected.

app/controllers/CompilerService.scala
```diff
@@ -203,8 +205,14 @@ object CompilerService extends Controller {
...
     (argMap get modelKey).fold(missingMsg.failureNel[String])(_.successNel[String])
 
-  private def fetchURL(url: String): ValidationNel[String, String] = {
-    Try(new URL(url)) map {
+  private def fetchURL(url: String, hostUri: String): ValidationNel[String, String] = {
+    val locallyHostedRegex = new Regex(s"^https?://$hostUri/assets/([A-Za-z0-9%/]+\\.nlogo)$$", "file")
+
+    Try {
+      locallyHostedRegex.findFirstMatchIn(url).flatMap(m =>
+          Assets.resourceNameAt("/public", m.group("file")).flatMap(Play.resource)
```
mrerrormessage: @TheBizzle , is this basically what you were looking for?

TheBizzle:
@mrerrormessage, in wake of these changes, have you observed the kind of performance improvements you were looking for?

mrerrormessage:
Yes. Without these changes, the tests will not even finish. With the changes, the tests finish and `/compile-url` is significantly faster.

Issue #212, opened 2015-03-16
IE: Patches are really blurry


TheBizzle:
Demonstration:

![badwsp](https://cloud.githubusercontent.com/assets/912890/6660533/be7c15f8-cb69-11e4-9aab-8065a07edc02.png)

This occurs in both IE10 and IE11.

qiemem:
Patches were originally done by using `drawImage` to scale the pixel data onto a canvas of the correct size. Due to [this bug](https://code.google.com/p/chromium/issues/detail?can=2&q=134040&colspec=ID%20Pri%20M%20Week%20ReleaseBlock%20Cr%20Status%20Owner%20Summary%20OS%20Modified&id=134040), this technique resulted in images strikingly similar to the one here. So I switched to stretching out the pixel canvas itself.

For #154, I'm switching back to scaling with `drawImage` (made possible by that bug being fixed). Hopefully, IE let's you modify the interpolation algorithm used by canvases. [Can I use](http://caniuse.com/#search=pixelated) says they have their own prefixed css attribute to accomplish this (though it's not clear to me if this refers just to `img` elements or to canvases as well).

qiemem:
Ah woops, looks like I switched the scaling techniques. [See this comment.](https://github.com/NetLogo/Galapagos/blob/a48724cc01f44e66a2fffdda411db580960e1c85/app/assets/javascripts/TortoiseJS/agent/view.coffee#L345)

So, I wanted to just scale with CSS, but that bug prevented me. So we're scaling with `drawImage` instead. [Looks like we just need to add an `ms` prefix to the `imageSmoothingEnabled` context property](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled#Browser_compatibility), like we do with [all the other browsers](https://github.com/NetLogo/Galapagos/blob/a48724cc01f44e66a2fffdda411db580960e1c85/app/assets/javascripts/TortoiseJS/agent/view.coffee#L375).

Again, I'll try to fix this with #154.

qiemem:
I've confirmed that

    @ctx.msImageSmoothingEnabled = false;

fixes the problem. I'll fix it with #154, since that's changing all the rendering code anyway.

qiemem:
[...]

PR #224, opened 2015-03-27
Add autoprefixer and fix IE

qiemem:
#212 is still an issue in IE10, but is fixed here in IE11.

Also monitors' output boxes don't stretch out in IE10, but are otherwise aligned correctly.

@TheBizzle Take a look and let me know how you'd like it squashed.

[Review on Reviewable](https://reviewable.io/reviews/netlogo/galapagos/224)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /.ensime_lucene/
 /.idea
 /logs/
+/node_modules/
 managed/
 project/project/
 target/
--- a/app/assets/javascripts/TortoiseJS/agent/view.coffee
+++ b/app/assets/javascripts/TortoiseJS/agent/view.coffee
@@ -2,6 +2,8 @@ class window.AgentStreamController
   constructor: (@container, fontSize) ->
     @view = new View(fontSize)
     @turtleDrawer = new TurtleDrawer(@view)
+    @patchDrawer = new PatchDrawer(@view)
+    @spotlightDrawer = new SpotlightDrawer(@view)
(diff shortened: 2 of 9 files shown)


TheBizzle:
Is there a particular reason that you [chose Autoprefixer 1.2.0](https://github.com/NetLogo/Galapagos/blob/1d78a7d1a79f2996f42e28931d9000d49d469b3c/package.json)?  [The latest version is 5.1.0](https://github.com/postcss/autoprefixer/releases).  My guess about the reasoning behind this choice is that maybe [SBT-Autoprefixer does not support more-modern versions of Autoprefixer](https://github.com/matthewrennie/sbt-autoprefixer/issues/3).

Also, [from the sound of it](https://github.com/matthewrennie/sbt-autoprefixer/issues/4), SBT-Autoprefixer is not being maintained.  I also find it super janky that we're requiring on the project through a URI to the latest state of a Git repo.  This ruins reproducibility of builds.

I can live with using SBT-Autoprefixer so we can get this out the door in a short amount of time, but I think we should replace it in short order.  I'm open to suggestions, but the most-obvious non-kludgy solution to this seems to me to be using a proper JavaScript build tool—preferably Gulp—for this stuff, [like we do in Tortoise](https://github.com/NetLogo/Tortoise/blob/master/node.sbt).

qiemem:
Yup, that's fine. I was just trying to do things the SBT-Webt Way (tm). I would be happy to use a real javascript build tool.

PR #228, opened 2015-04-22
put scalac on orange alert

SethTisue:


SethTisue:
with these changes, `ModelCollectionCompilerTest` fails, so that would need to be investigated before merging

TheBizzle:
The color-coded Homeland Security Advisory System is no longer in use, so I'm afraid that I won't be able to accept this pull request until you tell me where `scalac` is now placed on this thing:

Food plate thing

SethTisue:
delicious turtle soup = protein!

PR #230, opened 2015-04-29
Wip link labels

skorlir:
Labbbels. :shipit:

### DIFF
--- a/app/assets/javascripts/TortoiseJS/agent/drawshape.coffee
+++ b/app/assets/javascripts/TortoiseJS/agent/drawshape.coffee
@@ -5,22 +5,21 @@ class window.ShapeDrawer
   constructor: (shapes) ->
     @shapes = shapes
 
-  setTransparency: (ctx, turtleColor) ->
-    ctx.globalAlpha = if turtleColor.length > 3 then turtleColor[3] / 255 else 1
+  setTransparency: (ctx, color) ->
+    ctx.globalAlpha = if color.length > 3 then color[3] / 255 else 1
 
-
-  drawShape: (ctx, turtleColor, shapeName) ->
+  drawShape: (ctx, color, shapeName) ->
     ctx.translate(.5, -.5)
     ctx.scale(-1/IMAGE_SIZE, 1/IMAGE_SIZE)
-    @setTransparency(ctx, turtleColor)
-    @drawRawShape(ctx, turtleColor, shapeName)
+    @setTransparency(ctx, color)
+    @drawRawShape(ctx, color, shapeName)
     return
 
-  drawRawShape: (ctx, turtleColor, shapeName) ->
+  drawRawShape: (ctx, color, shapeName) ->
     ctx.lineWidth = LINE_WIDTH
     shape = @shapes[shapeName] or defaultShape
     for elem in shape.elements
-      draw[elem.type](ctx, turtleColor, elem)
+      draw[elem.type](ctx, color, elem)
     return
 
 class window.CachingShapeDrawer extends ShapeDrawer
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/TortoiseJS/agent/drawshape.coffee
```diff
@@ -8,16 +8,15 @@ class window.ShapeDrawer
   setTransparency: (ctx, turtleColor) ->
     ctx.globalAlpha = if turtleColor.length > 3 then turtleColor[3] / 255 else 1
 
-
-  drawShape: (ctx, turtleColor, shapeName) ->
+  drawShape: (ctx, turtleColor, shapeName, thickness = 1) ->
```
qiemem: Should `turtleColor` still be called "turtleColor" even though it's no longer just the turtle's color.

app/assets/javascripts/TortoiseJS/agent/view.coffee
```diff
@@ -203,108 +204,21 @@ class TurtleDrawer
...
+  drawLink: (link, turtles, wrapX, wrapY) ->
+    @linkDrawer.draw(link, turtles, wrapX, wrapY)
 
   repaint: (model) ->
     world = model.world
     turtles = model.turtles
     links = model.links
-    if world.turtleshapelist != @drawer.shapes and typeof world.turtleshapelist == "object"
-      @drawer = new CachingShapeDrawer(world.turtleshapelist)
+    if world.turtleshapelist isnt @turtleShapeDrawer.shapes and typeof world.turtleshapelist is "object"
```
qiemem: Think `typeof world.turtleshapelist is "object"` should be just `world.turtleshapelist?`. Don't know why I did it the other way before.

app/assets/javascripts/TortoiseJS/agent/view.coffee
```diff
@@ -369,3 +283,251 @@ class PatchDrawer
     if world.patcheswithlabels
       @labelPatches(patches)
 
+
```
qiemem: `Line` and `LinkDrawer` should definitely be in a different file.

app/assets/javascripts/TortoiseJS/agent/view.coffee
```diff
@@ -369,3 +283,251 @@ class PatchDrawer
...
+    midpointX = (@x1 + @x2) / 2
+    midpointY = (@y1 + @y2) / 2
+    [midpointX, midpointY]
+
+class LinkDrawer
+  constructor: (@view, @shapes) ->
+    directionIndicators = {}
+    for name, shape of @shapes
+      directionIndicators[name] = shape['direction-indicator']
+    @linkShapeDrawer = new ShapeDrawer(directionIndicators)
```
qiemem: My initial feeling is that this should inherit from `ShapeDrawer` instead, but I'm not sure there's a good enough reason to do so. Is there a reason why you didn't?

qiemem:
The link curves the wrong way in this model: https://gist.github.com/qiemem/438863ea4f4201f4bee3

Also, when links have any curve, they have rounded ends. This can easily be reproduced by sticking a circle of radius `thickness` at the turtles. This also removes the need for obnoxious link end logic.

I propose ditching the stupid link end logic and always using circles.

app/assets/javascripts/TortoiseJS/agent/view.coffee
```diff
@@ -369,3 +283,251 @@ class PatchDrawer
...
+  shouldWrapInDim: (canWrap, dimensionSize, cor1, cor2) ->
+    distance = Math.abs(cor1 - cor2)
+    canWrap and distance > dimensionSize / 2
+
+  adjustLinkEnds: (link, end1Cor, end2Cor, trigShift, minForExtending, maxForExtending) ->
+    diff = Math.abs(link.thickness * trigShift) / 2
+    if minForExtending < link.heading < maxForExtending
+      [end1Cor + diff, end2Cor - diff]
+    else
+      [end1Cor - diff, end2Cor + diff]
```
qiemem: This logic doesn't take into account curviness. In the model I posted, `ask turtles [ st ]`. You'll see the turtles are very off-center. This is because the above logic pushes the link ends to the sides by `thickness`.

Turn to the darkside. Just use circles.

app/assets/javascripts/TortoiseJS/agent/view.coffee
```diff
@@ -369,3 +283,251 @@ class PatchDrawer
...
+
+  calculateLineAngle: (x1, y1, x2, y2) ->
+    shortestX = Math.min(x1 - x2, x2 - x1)
+    shortestY = Math.min(y1 - y2, y2 - y1)
+    Math.atan2(shortestY, shortestX)
+
+  calculateComps: (x1, y1, x2, y2, size) ->
```
qiemem: Comment please

app/assets/javascripts/TortoiseJS/agent/view.coffee
```diff
@@ -369,3 +283,251 @@ class PatchDrawer
...
+          offsetSubline  = @getOffsetSubline(x1, y1, x2, y2, xOff, yOff)
+          isCurved       = curviness > 0
+          isMiddleLine   = line is lines[1]
+          hasLabel       = label?
+
+          [midpointX, midpointY] = offsetSubline.midpoint()
+
+          if isCurved
+
+            [controlX,  controlY]  = @calculateControlPoint(midpointX, midpointY, curviness, xcomp, ycomp)
```
qiemem: If you always calculate control points, then all the logic below can be collapsed, since with `curviness == 0`,  `controlX == midpointX` and `controlY == midpointY`. `calculateControlPoint` is simple enough that I doubt the performance hit is worth the code complexity of the branches below.

app/assets/javascripts/TortoiseJS/agent/view.coffee
skorlir: I think Links have more to them than what ShapeDrawer is intended for. Like, curviness doesn't apply to most cases where ShapeDrawer is used. Or even wrapping.

qiemem: I don't follow. I was suggesting inheriting from `ShapeDrawer`, not combining the two.

skorlir: Right, I was saying this isn't a ShapeDrawer. It contains a ShapeDrawer; it does draw shapes; it also does much more than that.

qiemem: Ah, to me that says it's an extension of `ShapeDrawer`, but no big deal. You can leave it as is.

skorlir:
@qiemem Updated to resolve all the issues/problems you brought up.

app/assets/javascripts/TortoiseJS/agent/linkdrawer.coffee
```diff
@@ -0,0 +1,221 @@
...
+
+    if linkShape['direction-indicator'].rotate
+      @view.ctx.rotate(shapeTheta)
+    else
+      @view.ctx.rotate(Math.PI)
+
+    # Magic numbers c/o Esther -- JTT, JAB 4/13/15
+    scalingFactor = 4
+
+    onePixelThickness = 1 / scalingFactor / scalingFactor
```
qiemem: wow, really?

app/assets/javascripts/TortoiseJS/agent/linkdrawer.coffee
```diff
@@ -0,0 +1,221 @@
...
+    [controlX, controlY]
+
+  drawSubline: ({x1, y1, x2, y2}, dashPattern, thickness, color, isCurved, controlX, controlY) ->
+    @view.ctx.beginPath()
+
+    @view.ctx.setLineDash(dashPattern.map((x) => x * @view.onePixel))
+    @view.ctx.strokeStyle = netlogoColorToCSS(color)
+    @view.ctx.lineWidth   = thickness
+
+    @view.ctx.lineCap = if isCurved then 'round' else 'square'
```
qiemem: Ooo how convenient.

qiemem:
Looks good. But, it doesn't look like direction indicators are scaling properly. Their thickness grows as thickness increases, but their actual scale doesn't, so they become hidden by the link pretty quickly.

https://gist.github.com/qiemem/438863ea4f4201f4bee3

skorlir:
@qiemem fixed the direction indicator scaling problem, good catch. Not super excited about the solution, but it does seem to work. @TheBizzle was remarkably unopinionated on the situation, but seemed to think it was good enough.

TheBizzle:
Is there an emoticon for being offended?

mrerrormessage:
:rage4: ?

TheBizzle:
Doom guy!

No, Jordan, *you* are the demons.

PR #231, opened 2015-05-08
Follow!

qiemem:
Everything should work: follow wrapping, spotlight, mouse stuff (including updating position when follow wrapping changes mouse screen position does not).

Waiting to squash till end of review.

### DIFF
--- a/app/assets/javascripts/TortoiseJS/agent/linkdrawer.coffee
+++ b/app/assets/javascripts/TortoiseJS/agent/linkdrawer.coffee
@@ -108,15 +108,7 @@ class window.LinkDrawer
     @view.ctx.restore()
 
   drawLabel: (x, y, labelText, color) ->
-    @view.ctx.save()
-
-    @view.ctx.translate(x - 3*@view.onePixel, y + 3*@view.onePixel)
-    @view.ctx.scale(@view.onePixel, -@view.onePixel)
-    @view.ctx.textAlign = 'end'
-    @view.ctx.fillStyle = netlogoColorToCSS(color)
-    @view.ctx.fillText(labelText, 0, 0)
-
-    @view.ctx.restore()
+    @view.drawLabel(x - 3 * @view.onePixel, y + 3 * @view.onePixel, labelText, color)
 
   draw: (link, turtles, canWrapX, canWrapY) ->
     if not link['hidden?']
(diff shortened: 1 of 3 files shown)


qiemem:
I should note: this does *not* include the great refactor I'd been working on. Decided it wasn't worth it, though the repeated wrapping logic still bugs me.

TheBizzle:
Please rebase onto `master`.

qiemem:
@skorlir This PR includes the refactor we talked about. [`usePatchCoordinates`](https://github.com/NetLogo/Galapagos/pull/231/files#diff-93679edcf7457bb5eee16a45da9d5870R96) should wrap drawing code that wants to draw in patch coordinates. [Turtle and link drawing now uses it for instance.](https://github.com/NetLogo/Galapagos/pull/231/files#diff-93679edcf7457bb5eee16a45da9d5870R96) The view is now untransformed by default, so [untransformation code is no longer necessary](https://github.com/NetLogo/Galapagos/pull/231/files#diff-93679edcf7457bb5eee16a45da9d5870R96).

TheBizzle:
@qiemem, is there more on the way, or is this ready for my review now?

skorlir:
@qiemem This is good, because I no longer have to do untransformations, but I can't use `usePatchCoordinates` because it transforms the view context, and I'm drawing the to drawingLayer. Is there any reason to not just refactor `usePatchCoordinates` to accept a context as its argument?

Using `usePatchCoordinates` only when stamping the canvas is no good (I went ahead and tried just to be sure); I end up with the exact problem I had before where I can't scale things correctly. I need to be able to use the transform for patch coordinates separately from the view's drawing context.

qiemem:
Good point, I'll tweak it.

skorlir:
Aacctually you might not want to. I tried that, and this had the scaling
problem I had before. I was able to get things working on wip-pen-trails
yesterday by reverse-engineering a "PatchCorsToPixels" function and using
that instead. It feels kinda weird to be passing the drawing context for an
invisible canvas to the view anyway, and `usePatchCoordinates` works well
in functions using the view canvas (although its name is a little
misleading). I dunno. Check out the branch and let me know what you think.

qiemem:
Hmm, okay. We should meet to talk about this scaling issue.

qiemem:
@TheBizzle Hold off until @skorlir and I talk.

qiemem:
So squashed.

PR #236, opened 2015-06-01
upgrade to Play 2.3.9 (released May 9, 2015)

SethTisue:


SethTisue:
see https://www.playframework.com/changelog for changes. none look problematic to me offhand, and some look perhaps beneficial

PR #244, opened 2015-06-09
Wip test monitors docked

mrerrormessage:
This is for Galapagos changes which enable https://github.com/NetLogo/Tortoise/pull/12 to work in Galapagos. I **do not** expect this PR to be merged without serious modification, but I wanted to have something that works to start the discussion off on the right foot.

Within Galapagos, the UI depends on the widgets, and tortoise code depends on the UI. This hasn't been a problem until now because we have simply passed the widgets to the browser as part of a separate data structure. However, as part of Tortoise#12, widgets are now compiled with the rest of the model and put into a variable: `var widgets = [...]`, so that the UI cannot be constructed before the model is evaluated. However, some code in the model assumes that the UI has been setup prior to it's evaluation, which causes problems. I'm pretty convinced, based on NetLogo-JVM's use of events in GUI loading code, that the problem is at least partially inherent to NetLogo.

There may be a number of solutions which I am not clever enough or familiar enough with the code to imagine, and I would really appreciate hearing them. I came up with these:
1) Continue to have a separate "widgets" object/array which is used to reconstitute the UI before the model is `eval`'ed.
2) Create some sort of event queue which temporally decouples model evaluation from UI initialization. This is the approach I took here for the plot events in particular, creating a proxy object which queued up plot events. The `PlotOpsManager` and `PlotOpsProxy` listen as the model is loaded and replay the events once the UI has been constructed.
3) Define a function in the model that is called after the model has been `eval`'ed which performs model setup vis-à-vis the GUI. Right now, we perform model setup by [sticking a bunch of commands at the end](https://github.com/NetLogo/Tortoise/blob/master/resources/test/dumps/AIDS.js#L290-L294), but there's no reason those commands couldn't live in a function which gets called after `eval` has read in the model and the GUI has been setup. This would require some modification in Tortoise, but if we're going this route, the sooner we make those modifications the easier it will be.

@TheBizzle , @qiemem, please comment and let me know your thoughts. Unless we can come to a solution before then, I hope we can discuss this (at least briefly) in CCL-Devel tomorrow.


SethTisue:
I can confirm that has always been a source of trickiness in JVM NetLogo. The Code and Interface tabs are mutually interdependent.

app/models/json/JsonConverter.scala
```diff
@@ -0,0 +1,29 @@
...
+      JsNull   => TortoiseNull,
+      JsObject => TortoiseObject,
+      JsString => TortoiseString }
+
+import play.api.libs.json.{ JsArray, JsBoolean, JsNull, JsNumber, JsObject, JsString, JsValue }
+
+object JsonConverter {
+  def toTortoise(playJson: JsValue): TortoiseJson = {
+    playJson match {
+      case JsObject(elems)             => TortoiseObject(fields(elems.map(t => (t._1, toTortoise(t._2))).toSeq: _*))
```
TheBizzle: `elems.toMap.mapValues(toTortoise)`

mrerrormessage:
Comment addressed. Once the version of tortoise is final, we'll need to update `build.sbt`.

TheBizzle:
This looks good.  It will be ready for merge once the Tortoise SHA is finalized.

PR #245, opened 2015-06-15
Deduplicate test assets

mrerrormessage:
This gets tests running again. I haven't tracked down the exact cause of the bug, but I'm pretty sure it has to do with autoprefixer trying to autoprefix css files pulled out of webjars. This should fix it, as long as our tests don't depend on assets.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,5 @@
+import com.typesafe.sbt.web.Import.WebKeys.webJarsDirectory
+
 name := "Galapagos"
 
 version := "1.0-SNAPSHOT"
@@ -40,4 +42,14 @@ resolvers += bintray.Opts.resolver.repo("netlogo", "NetLogoHeadless")
 
 resolvers += Resolver.file("Local repo", file(System.getProperty("user.home") + "/.ivy2/local"))(Resolver.ivyStylePatterns)
 
-pipelineStages in Assets := Seq(autoprefixer)
+pipelineStages in Assets += autoprefixer
+
+includeFilter in autoprefixer := Def.setting {
+  val webJarDir     = (webJarsDirectory in Assets).value.getPath
+  val testWebJarDir = (webJarsDirectory in TestAssets).value.getPath
+  new FileFilter {
+    override def accept(file: java.io.File) = {
+      file.getName.endsWith(".css") && ! (file.getPath.contains(webJarDir) || file.getPath.contains(testWebJarDir))
+    }
+  }
+}.value


mrerrormessage:
Hold off on merging this, I may have another (more correct) fix.

TheBizzle:
> I'm pretty sure it has to do with autoprefixer trying to autoprefix css files pulled out of webjars

That would make sense out of why the autoprefixing is such an obnoxiously-long task.

> This should fix it, as long as our tests don't depend on assets.

I think it's true at the moment that our tests don't depend on assets, but I expect that there will come a time when that isn't true.  What's reason that this constraint exists?  Would it be difficult to work around?

mrerrormessage:
I think the correct fix is just to have the deduplicator pick the appropriate css file. As an example:
```
[error] /Users/rgg284/IdeaProjects/Galapagos/target/web/public/test/public/lib/codemirror/theme/vibrant-ink.css
[error] from
[error] /Users/rgg284/IdeaProjects/Galapagos/target/web/web-modules/test/webjars/lib/codemirror/theme/vibrant-ink.css
[error] /Users/rgg284/IdeaProjects/Galapagos/target/web/web-modules/test/webjars/lib/galapagos/lib/codemirror/theme/vibrant-ink.css
```
I'm not sure which of these is correct. It's pretty easy to ensure this only deduplicates css files (`f => f.getName.endsWith(".css")`) and to differentiate which file is which (`f => f.getPath.contains("galapagos")`) but I just wasn't sure which one was the "correct" one. Looking at two of the duplicate files, I found one of them had different content - one additional line added where `::selection` was changed to `::-moz-selection`. Other than that, the files just had a comment at the end about source maps. If there is a *correct* stylesheet, and we know that our deduplication is selecting the correct one, we should be fine regardless of whether our tests depend on assets or not. My initial thought would be that choosing the one from the webjar (the one that had not been run through autoprefixer) is the right move, but you know a lot more about assets and webjars, so I'd appreciate if you made the call on this one. Just let me know and I'll touch up the deduplicator, then we should be good to go.

TheBizzle:
> Looking at two of the duplicate files, I found one of them had different content - one additional line added where `::selection` was changed to `::-moz-selection`

Right, so it sounds like the autoprefixer is running on the Webjar files, which I think we don't really want.  We want the non-prefixed Webjar CSS files.  Can't we just tell the autoprefixer not to process those?

mrerrormessage:
I think so. I'm working on that right now. I'll let you know if/when I make progress. This fix should at least allow you and others to continue running Galapagos tests in the meantime, even if we never merge it in.

mrerrormessage:
The latest change excludes we jar css files from being autoprefixed. It makes `sbt test` work for me locally, we'll see if it breaks on travis for some reason.

TheBizzle:
:tada::tada::tada:

Hooray for "Autoprefixing CSS" no longer taking ages to complete!

:tada::tada::tada:

PR #237, opened 2015-07-08
Recompiling a model throws away its custom shape information

mrerrormessage:
Originally NetLogo/Tortoise#145, reported by @ToonTalk.

See [this model](http://li425-91.members.linode.com:9000/tortoise#https://gist.githubusercontent.com/TheBizzle/8a04dd988d7c5986402f/raw/2dbf8b1987f958db5ea62c38e197f0ca588922e5/Garbage.nlogo) for a simpler demonstration of the problem.

My model provides a custom turtle shape called "thing".  If you press "setup" right after loading the page, the "thing" shape appears correctly.  If you open the "Code" tab and press "compile", though, "setup" starts to show you the turtles with the default turtle shape.  Once this is done, there is no going back, unless you reload the page.

I've dug around in the code a bit, and it looks to me like shape information simply isn't being persisted for some reason.  We don't notice this on models like Wolf Sheep Predation, because their shapes are all defaults shapes.

The "thing" shape is not included at all in the array of serialized shapes passed into the workspace constructor in the compilation response.  This suggests to me that Beak or maybe the Galapagos server itself is doing something wrong.

### DIFF
--- a/app/assets/javascripts/TortoiseJS/control/session-lite.coffee
+++ b/app/assets/javascripts/TortoiseJS/control/session-lite.coffee
@@ -150,7 +150,9 @@ window.compile = (source, model, commands, reporters, widgets,
     model: model,
     widgets: JSON.stringify(widgets),
     commands: JSON.stringify(commands),
-    reporters: JSON.stringify(reporters)
+    reporters: JSON.stringify(reporters),
+    turtleShapes: JSON.stringify(turtleShapes ? []),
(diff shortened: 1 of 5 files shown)


mrerrormessage:
The code on this branch maintains custom shape information between recompiles. I found the library models "Breed Procedures Example" and "Link Breeds Example" to be helpful in testing.

TheBizzle:
A+

Issue #221, opened 2015-03-16
SVG lines should only change in length when resized


TheBizzle:
Demonstration:

![screen shot 2015-03-16 at 1 46 42 pm](https://cloud.githubusercontent.com/assets/912890/6676910/6dcd3e6c-cbf9-11e4-8de4-9a54e2af9366.png)

This is demonstrated with the Koch Curve model.  The bottom image (JVM NetLogo) is doing the "correct" thing.  There is only one turtle in the image; its shape is a blue line with a red circle.  As the model runs, the size of the turtle (and, consequently, the size of its shape) change.  The red circle should be resizing, but the blue line should only change in length (not thickness).

That said, I don't think this is important enough to get into 1.0.

TheBizzle:
Fixed in 42870fffd350dbf8deabdf1d1c19f0d02d8b3f57

PR #260, opened 2015-07-13
Fix line widths

qiemem:
This ignores the shape cache. We can bring that back if necessary.


qiemem:
This also fixes #221.

Not sure why Travis is failing. I didn't change anything server-side.

TheBizzle:
Sometimes Travis just needs a little bit of this:

Fonzie hitting jukebox

qiemem:
Well then, I guess this is ready.

qiemem:
Take it for a spin, @TheBizzle. See how the performance feels to you. I know there's a significant performance hit (the degree depends on lots of factors: number of turtles (of course), shape, size, and so forth), but the frame skipping keep things tolerable at least.

To disable frame skipping, you can run `session.drawEveryFrame = true`

PR #42, opened 2015-07-15
use a linter and make it fail the build when problems are found

mrerrormessage:
`coffee -l` would have caught the "sucketURL" in https://github.com/NetLogo/Teletortoise/commit/87018f5f41671d09ee902cbc6f389aed7f4feac1 , yes?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,6 @@ language: scala
 scala:
   - 2.11.7
 script:
+  - "sbt scalastyle"
   - "sbt coffeelint"
   - "sbt test"
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,5 @@ language: scala
 scala:
   - 2.11.7
 script:
+  - "sbt coffeelint"
   - "sbt test"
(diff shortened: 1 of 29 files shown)


qiemem:
Ya, that's totally fine. Just want the PR to discuss code snippets.

mrerrormessage:
I'm removing the coffeelint check around operator spacing. I agree that having more flexibility in spacing around operators can result in clearer code.

mrerrormessage:
@qiemem , I've fixed all but three of the coffeelint errors, if you wouldn't mind having a look at them:
1. Lines in the Ractive template that are too long (it might make sense to do an inline disable for these).
2. An "Implicit braces" error, but I couldn't figure out where the braces were missing, as the error occurs on a constructor method.

TheBizzle:
> Two lines in the Ractive template that are too long (it might make sense to do an inline disable for these).

I highly encourage breaking line length rules where the constraint feels particularly artificial (e.g. in an HTML template).

mrerrormessage:
OK, when Travis passes this should be ready for review/merge.

PR #263, opened 2015-07-16
Static site

mrerrormessage:
- `/tortoise` can be used with a static backing server
- Automatic deploys to THE CLOUD
- When NetLogo code is posted to the server (as opposed to an nlogo file) it uses the key `"code"` instead of `"model"`, which allows requests to look the same as responses.

EDITED (per @TheBizzle 's suggestion):
:cloud: :cloud: :cloud: :cloud: :cloud: :cloud: :cloud: :cloud: :cloud: :cloud: :cloud: :cloud:


TheBizzle:
I appreciate the amount of THE CLOUD in this pull request, but I think it's important that we get a bit more of THE CLOUD into this PR before I can merge it.  Sorry to rain on your parade.

qiemem:
> When NetLogo code is posted to the server (as opposed to an nlogo file) it uses the key "code" instead of "model", which allows requests to look the same as responses.

Both request and response should use `model`. I chose `model` because it worked for all request types (request method distinguishes the type).

qiemem:
Oh, I see where you're coming from on the `code` thing. `code` is include in the `url` and `nlogo` responses because its the server that parses out the code... hmm... Well, whatever, I guess it's fine.

app/assets/javascripts/TortoiseJS/control/session-lite.coffee
```diff
@@ -78,14 +78,19 @@ class window.SessionLite
   exportnlogo: ->
```
qiemem: `exportNlogo`

qiemem:
lgtm!

PR #211, opened 2015-07-20
Add test models for Beak, widgets, renderer

mrerrormessage:
It would be nice if we had a model—or maybe even several models, but at least one—that allowed us to quickly and easily assess the bulk of the UI appearance stuff quickly at a glance.  Ideally, it would have one of each type of input—possibly even one for each variety of each type of widget.  Preferably, running the model would also make use of a wrapping topology, all types of agents, differently colored and shaped agents of all types, etc.

Having this would be especially nice for checking browser compatibility.  Right now, there's really no sensible way to open up a new browser and see if everything displaying correctly in it.  Similarly, it would be useful, going forward, for regression testing of the UI.

### DIFF
--- a/app/assets/javascripts/models.coffee
+++ b/app/assets/javascripts/models.coffee
@@ -1,12 +1,18 @@
 exports.bindModelChooser = (container, selectionChanged) ->
+
+  PUBLIC_PATH_SEGMENT_LENGTH = "public/".length
+
+  adjustModelName = (modelName) ->
+    modelName.substring(PUBLIC_PATH_SEGMENT_LENGTH, modelName.length)
+
   setModelCompilationStatus = (modelName, status) ->
-    $("option[value=\"#{modelName}\"]").addClass(status)
+    $("option[value=\"#{adjustModelName(modelName)}\"]").addClass(status)
 
   populateModelChoices = (select, modelNames) ->
     select.append($('<option>').text('Select a model...'))
     for modelName in modelNames
-      option = $('<option>').attr('value', modelName)
-        .text(modelName)
+      option = $('<option>').attr('value', adjustModelName(modelName))
+        .text(adjustModelName(modelName))
       select.append(option)
 
   createModelSelection = (container, modelNames) ->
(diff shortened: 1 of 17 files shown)


mrerrormessage:
This should be ready for review:
* models are available only in dev mode
* models are in directory `demomodels`
* models and what they test are as follows:

Model | What it tests
------- | -------
All Widgets.nlogo | different types of widgets, different configurations
Cache Tester.nlogo | caching of shapes (size, color, etc)
Clear Test.nlogo | Various clearing primitives, pen-trails, stamps
Mouse Test.nlogo | Mouse Primitives
Patch and World Size.nlogo | `patch-size`, `resize-world` (this model is currently not working on Galapagos, for reasons I can't quite figure out)
Perspective Primitives.nlogo | `follow`, `ride`, `reset-perspective` (`reset-perspective` doesn't seem to be working)
Plot Test.nlogo | Histograms, dot plots, line plots
Shapes and Link Shapes.nlogo | Custom shapes, link shapes, line thickness
Stop Test.nlogo | `stop`
Topology Test.nlogo | Test for different topologies, line plots using plotxy

qiemem:
> Cache Tester.nlogo

We don't actually use a cache anymore, so this name isn't exactly accurate.

> (`reset-perspective` doesn't seem to be working)

`reset-perspective` totally works... what's going wrong? Works fine on master and linode.

mrerrormessage:
Yes, `reset-perspective` does work. I had it in a button with `stop-inspecting`, which doesn't yet work on Tortoise. I'll update the model to split those out into two separate buttons.

TheBizzle:
There seems to be a lot of stuff that these models test that isn't yet implemented (which is good, in a way).  Could you make sure that 1.1-targeted tickets are opened for these (mis)behaviors?

mrerrormessage:
Yeah, I'll go through sometime later this week and open tickets. `stop-inspecting` should be added before 1.0, we're just waiting on a Tortoise update.

mrerrormessage:
Tickets have been opened.

PR #271, opened 2015-07-23
Users should be given some indication when a model is compiling

mrerrormessage:
Currently, models can take a long time to compile—especially on mobile platforms.  It would be nice for us to let the user know that something is actually happening while the compilation is taking place (to discourage them from reloading the page or doing other such self-defeating things).

For the time being, I'm alright with a pretty minimal solution to this problem, but I think some kind of modal "loading" animation would be nice at some point.


mrerrormessage:
The css spinner looks like:
screen shot 2015-07-23 at 11 35 01 am 
Thanks @frankduncan for the idea of making a turtle-spinner!

I left this as two commits for the time being because I thought it might be useful to compare the gif / css. When it's ready for merge, they can be squashed.

I wasn't able to get it to display the spinner while it was recompiling / running command center tasks, although when I paused it in the debugger, the animation did show up. It may be that the recompilations / compilations happened too quickly or that there's something funny going on with redraws.

TheBizzle:
If reasonably possible, I'd also prefer that the framerate of the animation be faster, with more turtles (6-8).

Also, I think this works a bit nicer if the animation isn't horizontally centered.  I found that changing the `#spinner`'s `left` from `40%` to `375px` was more comfortable, but I'm open to being told that this is a bad idea.  The reason I like this is that users are typically going to be looking at the left side of the screen when they do something that triggers this animation (because that's where the model is, and that's the part of the screen where you select new models), and then the animation will appear at the center of the screen, off to the right, in their periphery, and then go away, and they will have to look back to the left again.  Admittedly, it's a minor point.

Lastly, I'm a little wary of us using SVG for this, but maybe my fears are unfounded.  @qiemem, can you think of any problems associated with using SVG for this?

mrerrormessage:
Updated. Spinner now looks like:
screen shot 2015-07-24 at 11 30 58 am

PR #275, opened 2015-07-24
Model list should be curated in production

mrerrormessage:
this was useful during development, but will probably just confuse/confound users

### DIFF
--- a/app/assets/javascripts/models.coffee
+++ b/app/assets/javascripts/models.coffee
@@ -1,18 +1,32 @@
-exports.bindModelChooser = (container, selectionChanged) ->
+exports.bindModelChooser = (container, selectionChanged, currentMode) ->
 
   PUBLIC_PATH_SEGMENT_LENGTH = "public/".length
 
-  adjustModelName = (modelName) ->
+  adjustModelPath = (modelName) ->
     modelName.substring(PUBLIC_PATH_SEGMENT_LENGTH, modelName.length)
 
+  modelDisplayName = (modelName) ->
+    stripPrefix = (prefix, str) ->
+      startsWith = (p, s) -> s.substring(0, p.length) is p
+      if str.startsWith(prefix)
+        str.substring(prefix.length)
+      else
+        str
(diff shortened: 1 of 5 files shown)


TheBizzle:
Don't these changes render the status CSS classes and the icons pointless?  I don't think it's a bad thing if it does, but then those things should be cleaned up.

mrerrormessage:
I suppose they make "not_compiling" pretty pointless, but do we want to keep the "unknown" around, in case some models aren't compiled? I wasn't sure about whether we wanted to keep showing the "not_compiling" models in development or not, I can see pros and cons on both sides - what are your thoughts?

TheBizzle:
I hadn't thought about `unknown`.  You're right to point out that that should stick around (which means that `compiling` should, too).

Here's my ideal case:

All of the icons continue to live on only in Development Mode.  On the static pages that we distribute, there should be no icons, and only models that compile should be shown in the list.

mrerrormessage:
That makes sense. I was a bit hesitant to do that at first because it requires pretty different behavior in development mode, but I agree that seems like the best way to go. I'll modify.

app/assets/javascripts/models.coffee
```diff
@@ -2,17 +2,27 @@ exports.bindModelChooser = (container, selectionChanged) ->
 
   PUBLIC_PATH_SEGMENT_LENGTH = "public/".length
 
-  adjustModelName = (modelName) ->
+  adjustModelPath = (modelName) ->
     modelName.substring(PUBLIC_PATH_SEGMENT_LENGTH, modelName.length)
 
+  modelDisplayName = (modelName) ->
+    adjustedModelPath = adjustModelPath(modelName)
+    if adjustedModelPath.substring(0, 10) == "modelslib/"
```
Idloj: Isn't there a `startsWith` function?

TheBizzle: Unfortunately, support for that is first being added in ES6.  The sake of compatibility, our project targets ES5.

mrerrormessage:
This should be ready for re-review.

PR #277, opened 2015-07-26
Add export HTML to /tortoise page

qiemem:
Also adds `/web` and `/standalone` pages that may now be easily used as templates for either non-standalone pages. Loading urls from them is easy:

`/standalone?http://netlogoweb.org/assets/modelslib/Sample%20Models/Art/Follower.nlogo`

As iframes, you can pass nlogo contents to them via a `postMessage`. `/tortoise` does this. Finally, you can also insert the nlogo contents in the `#nlogo-code` element before the pages load. This is how the HTML export works.

Finally, I also made the spinner work a bit better. Since our recompile is now being done synchronously, the spinner never had a chance to come up before the recompile. I made it so the `startLoading` function could asynchronously run the loading process to give the spinner the opportunity to come up.

@TheBizzle, this is ready for review. @mrerrormessage you may want to glance at it as well since it adds another export method and touches the spinner.

### DIFF
--- a/app/views/simulation.scala.html
+++ b/app/views/simulation.scala.html
@@ -10,9 +10,11 @@
     @style(route("stylesheets/widgets.css"))
     @style(route("stylesheets/classic.css"))
     @style(route("stylesheets/netlogo-syntax.css"))
+    @style(route("stylesheets/spinner.css"))
   </head>
-  <body>
-    <div id="model-container"></div>
+  <body style="margin: 0px;">
+    @views.html.spinner()
+    <div id="model-container" style="display: inline-block;"></div>
   </body>
   @script(route("lib/filesaver.js/FileSaver.js"))
   @script(route("lib/markdown-js/markdown.js"))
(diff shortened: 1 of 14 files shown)


qiemem:
Ah, woops, missed the linting errors. Just a moment.

app/views/simulation.scala.html
```diff
@@ -0,0 +1,96 @@
...
+    @style(route("stylesheets/widgets.css"))
+    @style(route("stylesheets/classic.css"))
+    @style(route("stylesheets/netlogo-syntax.css"))
+    @style(route("stylesheets/spinner.css"))
+  </head>
+  <body style="margin: 0px;">
+    @views.html.spinner()
+    <div id="model-container" style="display: inline-block;"></div>
+  </body>
+  @script(route("lib/jquery/jquery.js"))
```
qiemem: So I finally caved. Play's javascriptRoutes works with jquery by default, and it made the ajax for the standalone request feel less hacky. We could swap it out for a minimal ajax library, or just write our own with the same signature as jquery's if you guys prefer.

qiemem:
Ready for review.

mrerrormessage:
Could you add the `/web` and `/standalone` routes to the list of routes to generate for the static site [here](https://github.com/NetLogo/Galapagos/blob/master/build.sbt#L67)?

qiemem:
For sure. Should create-standalone be removed as well?

mrerrormessage:
Yep, if you could take care of that as well, that would be awesome.

TheBizzle:
@qiemem, is this ready to go, or are you and @mrerrormessage still sorting out issues with scraping and domain names?

mrerrormessage:
We just had a call. I'm hoping to get the scraper stuff done pretty quickly today, allowing @qiemem to finish this up tomorrow.

qiemem:
@mrerrormessage, want to just do that stuff on this branch so it can be part of this PR?

mrerrormessage:
Sounds good. Once I'm done with the play-scraper modifications, I'll make any changes needed on this branch.

mrerrormessage:
@qiemem , we are now scraping absolute URLs. Is there anything else that needs doing?

TheBizzle:
This is pretty awesome.

Once the TODO item has been addressed, this should be ready to be merged.  I squashed a lot of the commits, so note that the branch has been force-updated.

qiemem:
Force pushed (to rebase onto master).

Moved stuff to its own CSS file and cleaned it up a bit. Made the default styling a little different just because it needed to be something, and I thought it worked better with the centered layout.

qiemem:
@TheBizzle This should be ready, though might want another squashin'.

TheBizzle:
I force-updated the branch again, and I think it would be ready, if not for a problem I just discovered.

Steps to reproduce:

  1. Open any model
  2. Export as HTML
  3. Open the exported file in your browser
  4. In this exported file, export as HTML

You will be met with an error message:

> Couldn't get standalone HTML: [object Object]

I don't have a firm opinion on whether or not exported HTML pages should be able to exported again, but I'm inclined to think that they should be able to, since the Code and Info content are still editable, in which case, this is a bug that should be fixed.

qiemem:
I certainly intended to allow them to export again, and tested it on a previous version. I'll take a look.

qiemem:
Alright, this should be good to go. Note that the standalone pages must be accessed via HTTP to use "Export HTML".

PR #279, opened 2015-07-28
Add syntax highlighting to command center

### LINKED ISSUE #274: Command Center input field should use monospace font
ideally it would syntax-highlight too, but in the meantime, just getting a monospace font would be a trivial change and would help newbies understand what the Command Center even is

qiemem:
Fixes #274


qiemem:
@TheBizzle This is ready for review.

Issue #269, opened 2015-07-21
Widget differences between JVM NetLogo, NLW


mrerrormessage:
Per https://github.com/NetLogo/Galapagos/pull/211. 
Most widgets look just about exactly like their NetLogo counterpart, but there are a few differences:
* note widgets with opaque backgrounds / different font colors
* vertical sliders
* Buttons with shortcut keys
* Buttons with agent types
![screen shot 2015-07-21 at 1 22 58 pm](https://cloud.githubusercontent.com/assets/143198/8809054/c4a7b7e2-2fab-11e5-9f6d-3d5b2fa7c7a9.png)
I'm not sure how many of these ommissions are intentional ("we're not putting this in NetLogo Web") vs unimplemented ("we haven't gotten around to it yet"). @qiemem , perhaps you could provide a bit more detail.

TheBizzle:
> * note widgets with opaque backgrounds / different font colors
* vertical sliders
* Buttons with shortcut keys
* Buttons with agent types

All of these ought to be supported at some point.

qiemem:
> note widgets with opaque backgrounds / different font colors

I had no idea. Not intentional. That said... I think I prefer the transparent backgrounds. Font colors should definitely be supported.

> vertical sliders

Intentionally de-prioritized. #171 

> Buttons with shortcut keys

Targeting 1.1. #262 

> Buttons with agent types

I think this works... We just don't show visually indicate. Maybe we should. And do a forever button icon while we're at it.

The other big differences I see:

- Multi-line widget names

Should be fixed. They should use ellipses like buttons do now.

- The "change" button on multi-line input boxes

I *really* want input boxes to remain editable without hitting change. That said, also having "change" to open up a larger editor for easier editing seems like a good idea to me.

- Hiding widget names for monitor

I didn't know this was possible. This should be supported.

Are there any others I'm not seeing?

TheBizzle:
> * Hiding widget names for monitor
>
> I didn't know this was possible. This should be supported.

I don't think that's intentional behavior.  Looks to me like it's just an artifact of abusing font size in JVM NetLogo.

mrerrormessage:
That's correct, JVM NetLogo isn't great about small font sizes. You can't actually shrink/grow the height of monitor widgets, except by changing the font size.

SethTisue:
> I think I prefer the transparent backgrounds

[...]

PR #280, opened 2015-07-28
Widget fixes

qiemem:
Not sure how exactly to bundle fixes for #269. These should probably be merged in on their own as the other fixes are somewhat more complicated.

### DIFF
--- a/app/assets/javascripts/TortoiseJS/agent/widgets.coffee
+++ b/app/assets/javascripts/TortoiseJS/agent/widgets.coffee
@@ -28,8 +28,9 @@ window.bindWidgets = (container, widgets, code, info, readOnly, filename) ->
     filename: filename,
     consoleOutput: '',
     outputWidgetOutput: '',
-    markdown: markdown.toHTML
-    convertColor: netlogoColorToCSS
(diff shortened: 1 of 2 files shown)


qiemem:
@TheBizzle I believe this is ready for review and merge. If there's any other widget fixes you'd like to see in here first, let me know. If you'd prefer the action keys thing in its own PR, I can do that too.

TheBizzle:
If you check out Minesweeper in 5e4709a55116db4f35f4d48e60fc3f659f7c0002 (the commit that adds "action keys"), you'll see that the text for the "setup" and "mark" buttons is left-aligned, where it should be centered like the text for "go" is.  The button text is properly centered for all buttons when using the code from the commit before.  @qiemem, please resolve this, and rebase onto `master` before resubmitting.

TheBizzle:
Note that I only seem to be able to reproduce this behavior in Opera (30.0.1835.125), but not in Firefox or Chrome.

qiemem:
I was also able to reproduce it Safari and, oddly, Opera after I updated
it. Should be fixed now.

PR #284, opened 2015-07-29
Update plots when widgets update instead of on every plot command

qiemem:
This provides a considerable performance increase, particularly noticeable when you push up the speed slider. Plotting Example is a good test model.


SethTisue:
good catch

TheBizzle:
Good change

PR #276, opened 2015-07-30
NLW Error Pages

mrerrormessage:
We can customize errors for any / all of these pages. We can build one error page and use it everywhere, or build pages for specific errors. A common practice I've seen is having a 404 page, then a separate page for all other errors, but we could also use just one page "Something went wrong..." or something like that.
screen shot 2015-07-24 at 12 33 52 pm

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -24,6 +24,11 @@ class Application extends Controller {
       Ok(views.html.info())
   }
 
+  def serverError: Action[AnyContent] = Action {
+    implicit request =>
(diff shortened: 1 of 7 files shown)


mrerrormessage:
This adds a catch-all error page. When https://github.com/NetLogo/Galapagos/pull/278 is ready, it can be integrated in with this. The reason that we didn't use the Play `ErrorHandler` is that we really don't want play to error while we're scraping this page. The template provided here could easily be integrated into the Play `ErrorHandler` later.

PR #288, opened 2015-07-30
Remove bootstrap, cleanup unused classes

mrerrormessage:
I did *not* redo the styling on `/create-standalone`, and I'm wondering whether we should remove the route for it. We can keep the template and controller method just so we can bring it back later, if we need to, but now that Bryan's export to HTML is ready, I'm not sure of it's value. At any rate, this removes bootstrap. The pages I tested were `/` and `/info`. I looked at `/create-standalone`, and it is usable, but the input areas look very different from before.


TheBizzle:
Thanks for doing this.  Surely this task was positively obnoxious.

> We can keep the template and controller method just so we can bring it back later, if we need to

Git will handle that for us.  I'll throw it out.  I don't really see a reason to keep it around, especially if continue to keep and `/compile-url` and `/compile-nlogo` around.

TheBizzle:
Oh, and I'd be willing to offer you a sainthood for killing off `obtain.sbt`.

qiemem:
`/save-url`, `/save-code`, and `/save-nlogo` should still work to.

PR #289, opened 2015-07-31
Replace platform-list with properties-list

Idloj:
This is no longer a platform list.


TheBizzle:
Merged in 143a6c4e39a7289591e1f30ee38aee5379f92a3c.  Thank you!

PR #290, opened 2015-07-31
Prevent future timeouts in Compiler Service Integration

mrerrormessage:
Increases the timeout slightly and maps through futures. I tested this in an Ubuntu VM and was getting consistent failures before this fix and consistent passes afterwards. My suspicion is that the play server takes a long time to "warm up" on the first request (lazy initialization doesn't help), but tends to run quickly once warmed up. I think it would be good to re-run this 5-10 times on Travis, just to ensure that it works.


TheBizzle:
10 Travis attempts.  10 Travis successes.

You're winner

mrerrormessage:
:tada:

PR #300, opened 2015-08-07
Wip adjust tabs

mrerrormessage:
Grows/shrinks, with some cool transitions (done on a separate commit for easy removal, if not desired).
screen shot 2015-08-07 at 10 39 27 am
screen shot 2015-08-07 at 10 39 21 am

### DIFF
--- a/app/assets/javascripts/TortoiseJS/agent/console.coffee
+++ b/app/assets/javascripts/TortoiseJS/agent/console.coffee
@@ -88,7 +88,8 @@ window.ConsoleWidget = Ractive.extend({
 
   template:
     """
-    <div class='netlogo-command-center netlogo-widget'>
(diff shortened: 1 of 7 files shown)


mrerrormessage:
@TheBizzle , made the requested corrections. It's working just fine in Safari, Chrome, FF (Safari required the `webkit-` prefix). IE11 works functionally, but the arrows don't appear on the right side (apparently svg data-uri's in css don't work in IE). I don't think this is too much of a problem, but if you disagree, I can work on it a bit more.

TheBizzle:
Can you just drop the arrows in as base64-encoding images, like the NetLogo icon is?

mrerrormessage:
yeah, I'll do that.

Idloj:
You can use Unicode arrows

PR #259, opened 2015-08-12
`stop` doesn't stop forever buttons

mrerrormessage:
The procedures themselves exit correctly, but the associated forever doesn't unpress and will just call the procedure again.


mrerrormessage:
Forever buttons now unbutton themselves when `stop` is called. Tested on Fire and the "Stop Test" model.

PR #305, opened 2015-08-14
Test

TheBizzle:
Ignore this.  Testing something....


TheBizzle:
This pull request is garbage!  I should be ashamed of myself!

SethTisue:
seems to me like it's enough to mark the commit red or green; adding a comment on the ticket too is redundant noise. nbd, but can that be turned off?

PR #312, opened 2015-08-18
Change markdown parser

mrerrormessage:
Elected to go with a sanitizing library instead, after reading [this wiki page](https://github.com/showdownjs/showdown/wiki/Markdown's-XSS-Vulnerability-%28and-how-to-mitigate-it%29).


SethTisue:
good catch

PR #315, opened 2015-08-19
Makes info page sections linkable

mrerrormessage:
Adds little "link" symbol beside headers on `/info`, and makes said headers into links.

screen shot 2015-08-19 at 8 58 57 am

### DIFF
--- a/app/views/ViewUtils.scala
+++ b/app/views/ViewUtils.scala
@@ -1,0 +1,12 @@
+// (C) Uri Wilensky. https://github.com/NetLogo/Galapagos
+
+package views
+
+object ViewUtils {
+
+  def idSafe(s: String): String = {
+    s.replaceAll(" ", "-")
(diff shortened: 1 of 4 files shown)


TheBizzle:
I've pushed up to this branch with some small changes that I made.  These will get squashed down before the PR is merged.

The PR, as submitted, is most of the way there, but I dislike the fact that, when you follow one of these links, the question gets obscured the by the top bar.  I'd like to the first thing below the top bar to be the question.  StackOverflow has [some suggestions](http://stackoverflow.com/questions/15139644/scroll-to-id-with-fixed-top-bar) for how to deal with this.

Idloj:
Can the headers themselves be links, instead of having a link icon next to them?

mrerrormessage:
I managed to get the effect by adding a `visibility:hidden` anchor above each header. I expect using javascript would make it harder / more complicated to get the scroll / position correct on page load, but if you would prefer that we use that approach, I'm willing to work on it, just let me know.

PR #316, opened 2015-08-20
Make Model Name More Visible

mrerrormessage:
The selection box makes the model name small and sometimes hides it altogether. Make the model name more central. 

One idea (may or may not end up being feasible):
att00001

Bonus points if you can click the title to convert it to a single-line text input that will save changes when the input loses focus, with the title changes being recognized by the "Export: NetLogo" and "Export: HTML" actions.


mrerrormessage:
Can be edited:

screen shot 2015-08-20 at 2 15 13 pm

Displayed prominently (length is just to show how much space):
screen shot 2015-08-20 at 2 14 53 pm

Also reflected in the `document.title`. Thanks for the suggestion @Idloj

app/views/simulation.scala.html
```diff
@@ -47,48 +48,55 @@
...
+    <script type="text/nlogo" id="nlogo-code" data-filename="model.nlogo"></script>
 
     <script>
       var modelContainer = document.querySelector("#netlogo-model-container");
       var index          = window.location.href.indexOf("?");
       var nlogoScript    = document.querySelector("#nlogo-code");
       var standaloneURL  = "@routes.Local.standalone.absoluteURL";
+      var pageTitle      = function(modelTitle) {
+        if (modelTitle != null && modelTitle != "") {
+          return "NetLogo Web: " + modelTitle;
```
Idloj: The convention is "Document Name - Software Name" (or in this case: "modelTitle - NetLogo Web").

PR #317, opened 2015-08-24
Generate and serve favicon.ico instead of favicon.png

mrerrormessage:
At the moment, we're serving a hand-generated .ico file that is not part of the build process and was put in as part of a stopgap last week. Although I don't expect that we'll change the ".ico" frequently, it shouldn't require manual intervention when we do. Unfortunately, we can't rely on browsers to use the `shortcut icon` tag in the html header (see [this informative page about favicons](https://github.com/audreyr/favicon-cheat-sheet#the-images)), so we must serve a favicon or else deal with a number of 404s.

Use of this PR does require imagemagick. Since travis has imagemagick by default, and it's easy to get on virtually any platform, I assumed that wouldn't be a problem. If it is, we can rethink.

I considered two other alternatives, which may be worth further exploration if this solution is unsatisfactory:

* Redirect `favicon.ico` to `favicon.png` (decided it was better to avoid extra requests whenever possible)
* Serve `favicon.ico` as a PNG file with content-type `image/png`. This is supported by some browsers, but perhaps isn't the best behavior to depend on, given that ".ico" is registered as being associated with the content-type `image/vnd.microsoft.icon`. 

Details of code changes:

- Convert favicon.png as part of asset pipeline
- Replace references to favicon.png with references to favicon.ico


SethTisue:
Depending on ImageMagick seems fine to me, but, one thing: if ImageMagick isn't installed, does the error message indicate what the problem is?

mrerrormessage:
I will look into that.

mrerrormessage:
An unambiguous error is now provided in the event that im4java is not able to locate ImageMagick.

PR #319, opened 2015-08-31
Changes to error reporting to support JVM "export to NetLogo Web"

mrerrormessage:
Adds a new view error component, which is a lightbox overlay (like the spinner) and displays errors in place of `window.alert`. The reason to switch is to allow links within the error (not possible using alert).

The component is the same regardless of the error situation, but there are basically four error cases where the look is somewhat distinctive:

1. Online Mode, error when loading model
2. Online Mode, error after model loaded
3. Exported Mode, error when loading model
4. Exported Mode, error after model loaded

Find screencaps of each situation below. 

1. Online Mode, error when loading model
screen shot 2015-08-31 at 10 18 37 am

2. Online Mode, error after model loaded
screen shot 2015-08-31 at 10 20 03 am

3. Exported Mode, error when loading model
screen shot 2015-08-31 at 10 20 20 am

4. Exported Mode, error after model loaded
screen shot 2015-08-31 at 10 21 27 am

### DIFF
--- a/app/assets/javascripts/TortoiseJS/control/session-lite.coffee
+++ b/app/assets/javascripts/TortoiseJS/control/session-lite.coffee
@@ -9,7 +9,7 @@ MAX_REDRAW_DELAY     = 1000
 REDRAW_EXP           = 2
 
 class window.SessionLite
-  constructor: (@widgetController, @alerter) ->
+  constructor: (@widgetController, @displayError) ->
     @_eventLoopTimeout = -1
     @_lastRedraw = 0
     @_lastUpdate = 0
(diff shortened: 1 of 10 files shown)


TheBizzle:
Excellent, thanks.

Idloj:
Is there a way to show %20 as a space?

mrerrormessage:
We definitely *could* decode the URL, but I think that makes the error message more confusing. We want to show people exactly the resource we've tried to request on their behalf and allow them to troubleshoot the problem. With the link url-escaped, they can easily paste it into their URL bar or the command-line to check what the problem is using `curl`.

Idloj:
In the browser address bar, you can enter URL without encoding it, you can just type whatever you want including spaces, symbols and what not.

PR #322, opened 2015-09-02
Restyle error alert box to foreground actual error

mrerrormessage:
Now looks like:

screen shot 2015-09-02 at 11 43 35 am

Depends on https://github.com/NetLogo/Tortoise/pull/171.

### DIFF
--- a/app/views/errorAlert.scala.html
+++ b/app/views/errorAlert.scala.html
@@ -8,8 +8,8 @@
         <div class="alert-text standalone-text">
             It looks like you're using NetLogo Web in standalone mode.
             <br />
-            If the above error is being caused by a missing primitive, we recommend a quick visit to
-                <a href="http://netlogoweb.org" target="_blank">NetLogoWeb.org</a>.
+            If the above error is being caused by an unimplemented primitive, we recommend a quick visit to
+                <a href="http://netlogoweb.org" target="_blank">NetLogoWeb.org</a>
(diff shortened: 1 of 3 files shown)


app/views/errorAlert.scala.html
```diff
@@ -8,8 +8,8 @@ <h3 id="alert-title">Error</h3>
         <div class="alert-text standalone-text">
             It looks like you're using NetLogo Web in standalone mode.
             <br />
-            If the above error is being caused by a missing primitive, we recommend a quick visit to
-                <a href="http://netlogoweb.org" target="_blank">NetLogoWeb.org</a>.
+            If the above error is being caused by an unknown primitive, we recommend a quick visit to
```
TheBizzle: I feel like "unknown" isn't the correct word for it.  Seems to me that it **is** known, or else they wouldn't be getting this error.  Perhaps it would be better to use the term "unimplemented" or "not yet implemented"?

mrerrormessage: I agree that unimplemented is clearer. I wanted to match the wording of the error exactly. I can go over to Tortoise and change "unknown primitive" to "unimplemented primitive", then change it to unimplemented here.

TheBizzle: That all makes sense.  I would be favor of that.

mrerrormessage:
Updated, Tortoise branch it depends on is still building:

screen shot 2015-09-02 at 12 40 50 pm

Issue #313, opened 2015-08-18
Multi-line text input boxes are single line


ToonTalk:
I made a string multi-line input box and got

![image](https://cloud.githubusercontent.com/assets/2138696/9327969/7ce314fe-459c-11e5-8740-652d4074e0ad.png)

with no way to enter multi-lines (other than \n)

Chrome Version 44.0.2403.155 m (Windows 7)

TheBizzle:
Related is my comment in #269:

> Multiline Inputs should support editing the text in a larger editor

TheBizzle:
Fixed by bfc0dec

PR #328, opened 2016-03-08
fix multiline inputbox issue

garzon:
To fix issue #313 

I find that I should make pull request to https://github.com/qiemem/TortoiseJS in [the wiki](https://github.com/NetLogo/Galapagos/wiki/TortoiseJS), however, it is out of date. So I make a pull request to here.

Now multiline inputbox is available(using `<textarea>`). However, there is still a problem that `widget.multiline` is always `true`.

### DIFF
--- a/app/assets/javascripts/TortoiseJS/agent/component/input.coffee
+++ b/app/assets/javascripts/TortoiseJS/agent/component/input.coffee
@@ -15,7 +15,13 @@ window.RactiveInput = Ractive.extend({
            style="{{dims}}">
       <div class="netlogo-label">{{widget.varName}}</div>
       {{# widget.boxtype === 'Number'}}<input type="number" value="{{widget.currentValue}}" />{{/}}
-      {{# widget.boxtype === 'String'}}<input type="text" value="{{widget.currentValue}}" />{{/}}
+      {{# widget.boxtype === 'String'}}
+        {{#if widget.multiline === false}}
+          <input type="text" value="{{widget.currentValue}}" />
+        {{else}}
+          <textarea rows="4">{{widget.currentValue}}</textarea>
+        {{/if}}
+      {{/}}
       {{# widget.boxtype === 'String (reporter)'}}<input type="text" value="{{widget.currentValue}}" />{{/}}
       {{# widget.boxtype === 'String (commands)'}}<input type="text" value="{{widget.currentValue}}" />{{/}}
       <!-- TODO: Fix color input. It'd be nice to use html5s color input. -->


TheBizzle:
Thanks for the PR.

I expected that a solution to #313 would involve adding a pop-up editor (like in desktop NetLogo), so the editing space isn't restricted to the space of the widget.  I think that still *will* have to happen at some point (for which I will open another ticket), but this is an improvement over the current state of things, so I guess that I'm alright with it.

However, there are still a few things here that I'd like to see addressed before accepting it:

  * This PR adds a call to `widget.value.replace` for all Input widgets.  However, `widget.value` does not necessarily have a `replace` method.  It only has `replace` if that Input handles strings.  If it is (for example) a numerical Input, then `widget.value.replace` will be `undefined`.  And then the entire page will fail to work, because a JavaScript exception is thrown when trying to treat `undefined` like a function.  All types of Inputs must avoid throwing errors like this.
  * This PR currently makes *all* Inputs multiline.  This is not really your fault—it was a bug in NetLogo/NetLogo-Headless, which I've now fixed—but you'll need to rebase your changes onto the latest version of the Galapagos `master` branch to avoid this problem.
  * The `"NIL"` check is a great and necessary improvement.  However, since it's a change that's separate from (and not really related to) whether or not multiline Inputs are supported, I would like to see the `"NIL"` fix alone in a separate commit.

garzon:
@TheBizzle 
Thanks for your advice.
I decide to split the commit into two PRs since they are independent.
And I check the type of `widget.value` now to avoid the problem.

However, when I run the newest code on the master of NetLogo/Galapagos, I find that the browser can not fetch the entire tortoise-compiler.js and/or netlogo-engine.js and it complains that:
```
tortoise-compiler.js:3189 Uncaught SyntaxError: Unexpected end of input
netlogo-engine.js:2104 Uncaught SyntaxError: Unexpected end of input
```
The file stream interrupted in different positions, resulting in "unexpected end of input". No matter how many times I refresh the page.

I think there is some problems with the following code:
```
in app/controllers/CompilerService.scala
@@ -183,7 +183,7 @@ private[controllers] trait CompilationRequestHandler extends RequestResultGenera
       environment.resourceAsStream(fileName).map(tortoiseJs =>
         Result(
           header = ResponseHeader(OK, Map(CONTENT_TYPE -> "text/javascript")),
           body   = Streamed(StreamConverters.fromInputStream(() => tortoiseJs), None, None)
         )).getOrElse(NotFound)
     }
```

garzon:
Yes, there is no problem running the web app now. :)
I have updated this PR and now multiline inputbox should be available without throwing exceptions.

However, there is still a small problem that, in a single line inputbox, backslashes display as `\\`, even though I add these calls of `replace` function in `fillOutWidgets` in this PR and it produces the correct HTML code like this: `<input type="text" value="i am a single\line inputbox." autocomplete="off">`.
It is because there are bidirectional bindings of `widget.currentValue` and the "global" variables in `world.observer._varManager`, and the initial code generated automatically calls `world.observer.setGlobal` after `fillOutWidgets` is called. The initial code is something like this:
```js
....
var plotManager = workspace.plotManager;
var world = workspace.world;
var procedures = (function() {
  
  return {
  
  };
})();
world.observer.setGlobal("single", "i am a single\\\\line inputbox");   // setGlobal is called here
```
And here is the code related to the bidirectional bindings:
```js
in widgets.coffee, line 132:
  ractive.observe('widgetObj.*.currentValue', (newVal, oldVal, keyPath, widgetNum) ->
    widget = widgetObj[widgetNum]
    if world? and newVal != oldVal and isValidValue(widget, newVal)
      world.observer.setGlobal(widget.varName, newVal)
  )
line 196:
        if widget.varName?
          widget.currentValue = world.observer.getGlobal(widget.varName)
```
So, the `widget.currentValue` that set in the `fillOutWidgets` in this PR is overwritten by the value stored in the `_varManager`, after `updateWidgets` is called. As a result, the text in the inputbox displays as `i am a single\\line inputbox`, same as the value in the initial code, rather than the correct replaced one in `fillOutWidgets`, `i am a single\line inputbox`. In a word, `widget.currentValue` in `fillOutWidgets` does not matter at all.

So I think a good solution is, do not generate the initial code that calling `world.observer.getGlobal` for the value of UI widgets. The initial works should be done by `fillOutWidgets`.

garzon:
Or is there a better solution now? :)

TheBizzle:
I've been a bit busy lately, but I'm looking into it and will get back to you soon.

TheBizzle:
Alright, I've now updated `master` to use a version of the compiler that won't require us to do any client-side escaping/unescaping of newlines, `"NIL"`, etc.  You should be able to proceed now without worrying about that stuff.  Sorry that took so long.

By the way, Play 2.5 has been reintroduced.  It's not causing problems for me (or for the service that scrapes the server to turn it into static HTML pages for the website), but if you run into issues with it again, let me know, and we'll try to figure that out.  Either way, your changes should be mergeable if you rebase onto `master` and remove the changes you made to `widgets.coffee`.

garzon:
Now I have removed the changes in widgets.coffee.

However, same situation with Play 2.5 in Chrome, Version 49.0.2623.87 m (64-bit) (in the last time it was Chrome 48).
It seems like this a problem of Chrome. And there is no problem running the web app in Firefox.

TheBizzle:
Thanks for the PR.  Sorry it took so long to get this through.

The "connecting to a server" problem with play is interesting, and I _am_ able to reproduce that.  I'll look into it.

garzon:
Please inform me about that if you solve the problem.

By the way, if you have time, could you please review my proposal of GSoC 2016? It is about implementing Model-Making Tools in NetLogo Web.
Thanks! :)

PR #331, opened 2016-03-17
Make view change size to match world and make model container dynamically adjust size

qiemem:


TheBizzle:
I'm really glad that you got this working with a pretty small set of code changes.

However, there's a minor thing that I'd like to see addressed before merging it, which is that, if you shrink down the View, you can get some weird horizontal scroll bar/container-sizing issues.  Below is a visual demonstration of this with Wolf Sheep Predation, where all I had to do to trigger the bug was run `set-patch-size 3`.

evilscrollbar

evilscrollbar2

Note that there's an awkward horizontal scrollbar that must be used in order to access the "Export:" buttons, which are out of the viewport.  Also, note that the "speed" slider and the accordion are centered as if the View is the right-most boundary.  This leads me to believe that the viewport's size is being set by looking at the widget container's size, and my guess about the solution is that `.netlogo-header` needs to be resized along with the widget container.

qiemem:
Fixed!

TheBizzle:
Awesome, thanks!

PR #360, opened 2017-03-22
Added "File/New" nlogo file to models library

CamdenClark:
Very trivial addition. This can be turned into a button or similar in the future, but simply having a "File/New" at the top of the models library could be very useful for users and is some kind of starting point at the very least.

I just now got around to having my head around the repository and wanted to commit something.

I signed the CLA.

Tests: 
`[info] ScalaTest
[info] Run completed in 10 seconds, 310 milliseconds.
[info] Total number of tests run: 12
[info] Suites: completed 3, aborted 0
[info] Tests: succeeded 12, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.`

### DIFF
--- a/app/assets/javascripts/plot/highchartsops.coffee
+++ b/app/assets/javascripts/plot/highchartsops.coffee
@@ -92,7 +92,12 @@ class window.HighchartsOps extends PlotOps
     super(resize, reset, registerPen, resetPen, addPoint, updatePenMode, updatePenColor)
     @_chart              = Highcharts.chart(elemID, {})
     @_penNameToSeriesNum = {}
-
+    #These pops remove the two redundant functions from the export-csv plugin
+    #see https://github.com/highcharts/export-csv and
+    #https://github.com/NetLogo/Galapagos/pull/364#discussion_r108308828 for more info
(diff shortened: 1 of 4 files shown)


TheBizzle:
Yeah, so, seeing this in action now, I've decided that don't like the user experience for it yet.  I'm glad you took this up, and I'm glad that it works, but I think we should tweak some things before merging it.

  [ ] `File/New` should be removed from the Models Library.  I think it's probably too roundabout and unexpected for most people.
  [ ] Below the "Export" stuff in the top-right corner, we should put a "New" button.  Clicking it opens a dialog that asks, "Are you sure you want to open a blank model?"  If "no", stay.  If yes, open a blank model.  The "open a blank model" thing might still benefit from using this `New.nlogo` file as a template for what is sent to compiler to kick things off, but the text for that template should live in the CoffeeScript files somewhere.
  [ ] Rather than the option being accessible at the URL hash `#<baseURL>assets/modelslib/File/New.nlogo`, I think that special-casing the hash `#new` would be better for activating this (and should function the same as clicking the "new" button does)

CamdenClark:
Working on this on a new branch

PR #361, opened 2017-03-24
Export-csv xls functionality to highcharts plots

CamdenClark:
This commit allows users to export the data inside the highcharts plots to CSV and XLS. I adapted it from https://github.com/highcharts/export-csv/blob/master/export-csv.js which is MIT licensed. It's my understanding that MIT is compatible with GPL, please correct me if I'm wrong.

It works surprisingly seamlessly on my end. Only adds about 10KB compiled and the file loads within 3ms locally. Only loads once per simulation, right after the base highchart class is loaded.

`[info] ScalaTest
[info] Run completed in 10 seconds, 207 milliseconds.
[info] Total number of tests run: 12
[info] Suites: completed 3, aborted 0
[info] Tests: succeeded 12, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[info] Passed: Total 19, Failed 0, Errors 0, Passed 19
`
All tests pass. Thanks again!

### DIFF
--- a/app/assets/javascripts/plot/export-csv.coffee
+++ b/app/assets/javascripts/plot/export-csv.coffee
@@ -1,0 +1,301 @@
+###*
+# A Highcharts plugin for exporting data from a rendered chart as CSV, XLS or HTML table
+#
+# Author:   Torstein Honsi
+# Licence:  MIT
+# Version:  1.4.7
+###
+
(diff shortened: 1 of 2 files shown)


CamdenClark:
Wanted to get fix the coffeelint errors.

PR #362, opened 2017-03-24
 Export-csv xls functionality to highcharts plots (Fixed coffeelint errors)

CamdenClark:
This commit allows users to export the data inside the highcharts plots to CSV and XLS. I adapted it from https://github.com/highcharts/export-csv/blob/master/export-csv.js which is MIT licensed. It's my understanding that MIT is compatible with GPL, please correct me if I'm wrong.

It works surprisingly seamlessly on my end. Only adds about 10KB compiled and the file loads within 3ms locally. Only loads once per simulation, right after the base highchart class is loaded.

[info] ScalaTest [info] Run completed in 10 seconds, 207 milliseconds. [info] Total number of tests run: 12 [info] Suites: completed 3, aborted 0 [info] Tests: succeeded 12, failed 0, canceled 0, ignored 0, pending 0 [info] All tests passed. [info] Passed: Total 19, Failed 0, Errors 0, Passed 19
All tests pass. Thanks again!


app/assets/javascripts/plot/export-csv.coffee
```diff
@@ -0,0 +1,309 @@
...
+  if seriesTypes.map
+    seriesTypes.map::exportKey = 'name'
+  if seriesTypes.mapbubble
+    seriesTypes.mapbubble::exportKey = 'name'
+  if seriesTypes.treemap
+    seriesTypes.treemap::exportKey = 'name'
+  return
+)
+# ---
+# generated by js2coffee 2.2.0
```
TheBizzle: Functionally, this PR is awesome!  My only complaints are about how we're going about it.  Namely, instead of dropping an outside file into the repo, our standard way of dealing with this sort of thing—for a number of reasons that I can go into more at length, if you want—is to depend on the external file through our package management system.

According to webjars.org , we can depend on this file in SBT by adding the entry `"org.webjars.bower" % "github-com-highcharts-export-csv" % "1.4.3"` to `build.sbt`.  0a5b2b73e6ebdd8a5350bdc9d6d86cb8f841429a is an example of a commit where something similar was done, in that we added a dependency to `build.sbt` and then specified that it should be included in NLW by adding lines to `simulation.scala.html` and `CompilerService.scala` to include the file by its URL.

With WebJars, it's sometimes a bit of a challenge to guess what URL we'll need to provide to get this file out of the WebJar.  The WebJars site says that the file is at `META-INF/resources/webjars/github-com-highcharts-export-csv/1.4.3/export-csv.js`.  As a result, I would guess that the URL we should use try to get it from is `lib/github-com-highcharts-export-csv/export-csv.js`.  Try that and see if it works.

CamdenClark: Awesome! I didn't even realize we were using bower yet, but that definitely makes more sense. I'll get to work on this. Thanks for reviewing these quickly.

PR #364, opened 2017-03-27
Adding export csv and xls functionality to plots through webjars

CamdenClark:
Sorry for yet another PR, but I was having trouble rebasing on that current branch. So I just created another branch and rebased the commits. (referencing: https://github.com/NetLogo/Galapagos/pull/362)

This implements the exportcsv and xls functionality using webjars. The package adds a menu item that also displays an HTML table that deletes the data. I added a line to highchartsops that pops that menu item off (because it's useless functionality).

I hope this solves all the issues. I've learned a ton going through this process, thanks!

### DIFF
--- a/app/assets/javascripts/plot/highchartsops.coffee
+++ b/app/assets/javascripts/plot/highchartsops.coffee
@@ -93,7 +93,10 @@ class window.HighchartsOps extends PlotOps
     @_chart              = Highcharts.chart(elemID, {})
     @_penNameToSeriesNum = {}
 <<<<<<< HEAD
-    #These pops remove the two redundant functions from the export-csv plugin see https://github.com/highcharts/export-csv and https://github.com/NetLogo/Galapagos/pull/364#discussion_r108308828 for more info
+    #These pops remove the two redundant functions from the export-csv plugin 
+    #see https://github.com/highcharts/export-csv and 
+    #https://github.com/NetLogo/Galapagos/pull/364#discussion_r108308828 for more info
+    #--Camden Clark (3/27/17)
 =======
 >>>>>>> aba93493332b3c78930b9f6efc4c2e75c8615ae9
     @_chart.options.exporting.buttons.contextButton.menuItems.pop()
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/plot/highchartsops.coffee
```diff
@@ -92,7 +92,7 @@ class window.HighchartsOps extends PlotOps
     super(resize, reset, registerPen, resetPen, addPoint, updatePenMode, updatePenColor)
     @_chart              = Highcharts.chart(elemID, {})
     @_penNameToSeriesNum = {}
-
+    @_chart.options.exporting.buttons.contextButton.menuItems.pop()
```
TheBizzle: Hmm.  This is super-weird.  On my end, I'm actually seeing that popping once is even enough.  I need to do it twice to get rid of the duplicate entries.  Do you know why this is happening?

If it's justified—and given the small number of lines of code we had to add in order to get this functionality, I'm thinking that it very likely *is* justified in some way or another—we should add the second `pop` (or just drop the last two items with `slice`), and comment this.

CamdenClark: @TheBizzle I thought it might be useful for users to have access to exporting the data to XLS but that might be a bit redundant. 

The export-csv plugin for highcharts adds three functions to the highchart menu

1. Download CSV - self-explanatory
2. Download XLS - self-explanatory
3. Show table - basically just displays an HTML table of the data, this is the most useless because it ruins the formatting and just gets in the way.

I'll add an extra pop to get ride of the Download XLS so we only have the CSV functionality.

TheBizzle: Ohhh, I didn't realize that the three options were actually different.  The tooltip went off the screen for me, and both `.csv` files and `.xls` files open with the same program on my computer, so they seemed the same.  This is making more sense.  I have way less of a strong preference now, as to whether or not the XLS option should be shown, but I guess we can keep it off.  It seems a little redundant.

But can you make sure to leave a comment there next to the `pop`s, explaining why they're being done?

CamdenClark: Yes, I've made such a mess of this PR, I apologize. I'm going to rebase to a single commit.

Issue #298, opened 2015-08-06
Link arrows way too big in some cases


qiemem:
A good example is Diffusion on a Directed Network:

NLW:
![image](https://cloud.githubusercontent.com/assets/28215/9114536/0829068a-3c20-11e5-8c31-ce48b98ffc37.png)

Desktop:

venky18:
@TheBizzle I would like to work on this can you guide me?

TheBizzle:
Sure.

The line that's responsible for drawing this in NetLogo Web is [this one](https://github.com/NetLogo/Galapagos/blob/master/app/assets/javascripts/TortoiseJS/agent/linkdrawer.coffee#L156).  The relevant code for this in desktop NetLogo is [here](https://github.com/NetLogo/NetLogo/blob/hexy/netlogo-core/src/main/shape/LinkShape.scala#L157-L158).

Let me know if there's anything else you would like to know.

TheBizzle:
Here's how the desktop NetLogo version of Link Breed Example looks:

![desktop-normal](https://cloud.githubusercontent.com/assets/912890/24722443/5e200926-1a09-11e7-9d00-1f4b9d0f004a.png)

Here is how NetLogo Web's version of it currently looks:

![nlw-normal](https://cloud.githubusercontent.com/assets/912890/24722383/2a6d78d4-1a09-11e7-8ec6-eb1a42b3d3c8.png)

So our indicators are far, far too large, and a bit too thick.

It might seem that the obvious solution to this problem is to mess with the coefficient of scaling [here](https://github.com/NetLogo/Galapagos/blob/58ce4ca1f1145c88d7bc9f91561a46367f7a3936/app/assets/javascripts/TortoiseJS/agent/linkdrawer.coffee#L98-L101).  However, that alone is not sufficient.  Here is what the model looks like if we change the `2` to `15`:

![nlw-reduced](https://cloud.githubusercontent.com/assets/912890/24722382/2a6cedc4-1a09-11e7-8429-bcd5310d340b.png)

The lengths of the indicators are now roughly correct, but they're very hard to see, because our thickness got scaled down into oblivion.

For all the faults of the current implementation, the indicators it produces are at least easily visible.  Any solution to this needs to find a reliable way of producing both indicators that have both a suitable length and a suitable thickness.

venky18:
@TheBizzle what is the Galapagos equivalent of `numLines` and `cellSize`? I read both files but couldn't find what these 2 keyword stands for?

[...]

PR #366, opened 2017-04-05
Fixing "Link arrows way too big in some cases" #298

syedshariq-a:
This is a temporary fix for the link arrows too big. I am still working on it to make the arrows better. However these changes fix the issue mentioned in #298.


TheBizzle:
I don't think that it's a good solution to this problem to change only the coefficients of scaling.  I've commented on #298 with my thoughts on this, but, for example, here's a picture of Link Breeds Example (for comparison with the others), using the scaling values from your PR:

nlw-4

The indicators are smaller, yes, but the down-scaling has caused them to look jaggy and unpleasant.  It's not apparent to me that these indicators are ultimately better than the ones we currently have or closer to how we want them to look.

PR #367, opened 2017-04-20
Adding New model button

CamdenClark:
@TheBizzle 

This is my current implementation for this problem. I just add a New.nlogo file to the models library folder (which is untracked by the drop-down select box because I didn't edit the scala models file). This is used as a template when you click the button.

When you click the button, a confirm box prompts the user letting them know it will create a new model and that their changes will be lost.

This is the easiest (and hackiest) way to approach this as we'd be adding a new nlogo file to the models library. It's easier because it uses the openURL JS method already created in the tortoise.html.scala file so I don't have to paste an entire nlogo file into a JS file somewhere.

Thoughts?


TheBizzle:
Sometimes Travis needs the ole Fonzy treatment.  I've restarted the build.

CamdenClark:
Okay, new commit with template dumped into coffee file that exports the template variable. Include it next to the models.js in the tortoise.scala.html file. Works for me, all tests passed on my end. Let me know if I should fix any styling issues.

TheBizzle:
This got merged as 2544a7a976f668af20875cf049fda592618311ee .  Thanks for your contribution!

PR #368, opened 2017-05-23
Input box template should not bind multiline value for editing on non-String input types

LaCuneta:
This change is actually to fix [Tortoise issue 182 about error messages when using the command center](https://github.com/NetLogo/Tortoise/issues/182).


TheBizzle:
@LaCuneta, there's a bunch of extra stuff in this PR because your work was based off of the `wip-play-2.6` branch, and you're requesting to merge your work into `master` here.

To fix this, either aim the PR at `wip-play-2.6`, or do the work on top of `master` (since it turned out that you didn't need to update NetLogo `.jar`s, anyway) and then keep your PR aimed at `master`.

app/assets/javascripts/TortoiseJS/agent/component/input.coffee
```diff
@@ -127,8 +127,13 @@ window.RactiveInput = RactiveWidget.extend({
     """
     {{>input}}
     {{>contextMenu}}
+    {{# widget.boxedValue.type === 'Color' || widget.boxedValue.type === 'Number' }}
     <editForm idBasis="{{id}}" boxtype="{{widget.boxedValue.type}}"
+              display="{{widget.display}}" />
+    {{ else }}
+        <editForm idBasis="{{id}}" boxtype="{{widget.boxedValue.type}}"
               display="{{widget.display}}" isMultiline="{{widget.boxedValue.multiline}}" />
+    {{/}}
```
TheBizzle: It occurred to me last night (and I meant to tell you then) that this solution is a bit non-optimal.  We can avoid the code duplication here by doing something like what we do [here](https://github.com/NetLogo/Galapagos/blob/master/app/assets/javascripts/TortoiseJS/agent/component/editor.coffee#L23), where we just surround the optional property in an `if`.

LaCuneta: Hey Jason, thanks for the tip, I can definitely do it that way.  And I actually did try to get the work pointed back at master before pushing it up for the PR, but that obviously didn't work.  I'll get that resolved, too.

LaCuneta:
Closing to re-open with a clean branch off of master.

PR #369, opened 2017-05-23
Input box template should not bind multiline value for editing on non…

LaCuneta:
…-String input types

This change is actually to fix [Tortoise issue 182 about error messages when using the command center](https://github.com/NetLogo/Tortoise/issues/182).  Updated the change to be for the input box multiline property only, per the comments on the [previous PR off of the wrong branch](https://github.com/NetLogo/Galapagos/pull/368).  I tested the change manually in Firefox and Chrome and no longer receive the input box error when using the command center, as expected.


TheBizzle:
This was merged as 3da9d93a7f48a88a8d237acbdf96e253342660ba .  :tada:

PR #378, opened 2018-02-26
Fixing #298

kunal-mohta:


kunal-mohta:
Yes, you are right. But I think this solution provides a bit of an improvement over the previously suggested solutions. 
The following are the screenshots of the current link arrows - 
screenshot 2018-02-26 23 13 17
screenshot 2018-02-26 23 13 42
screenshot 2018-02-26 23 14 10

And the following are after using the solution in this PR - 

screenshot 2018-02-26 23 11 32
screenshot 2018-02-26 23 07 45
screenshot 2018-02-26 21 07 44

I feel like it is a drastic change, maybe temporary, till you find a better, more compatible solution.
What do you think?

app/assets/javascripts/TortoiseJS/agent/linkdrawer.coffee
```diff
@@ -95,12 +95,13 @@ class window.LinkDrawer
...
 
     if thickness <= 1
-      scale         = 1 / @view.onePixel / 2
+      scale         = 1 / @view.onePixel / 7
       realThickness = thickness * baseThickness
     else
       scale         = thicknessFactor / 2
       realThickness = baseThickness
 
+    ctx.lineWidth = 10
```
LaCuneta: I'm not a big fan of setting the context width directly here, since that looks like it should be the job of `@linkShapeDrawer`, and that's why we're going to the trouble of calculating a `realThickness` value and passing it in.  

I tried setting a different `realThickness` value first by multiplying a constant on line 99, then by passing in hard-coded values directly, and didn't see a difference in link arrow size.  Turns out that because the `ShapeDrawer` takes in a `@onePixel` field in [the constructor](https://github.com/NetLogo/Galapagos/blob/staging/app/assets/javascripts/TortoiseJS/agent/drawshape.coffee#L4), but no value is being passed in [here](https://github.com/NetLogo/Galapagos/blob/staging/app/assets/javascripts/TortoiseJS/agent/linkdrawer.coffee#L14).  That left a default line width value being used, which is why it can appear so thin.

So I think I'd like to see the following modifications: 

- Remove the direct setting of the context `lineWidth`.
- Fix the existing bug where the `@view.onePixel` value is not passed to the `@lineShapeDrawer` constructor, so it'll actually use the `realThickness` value that is passed in. 
- Add a minimum value check for the `realThickness` value (say, `0.5`) before it's passed to the `@linkShapeDrawer` to draw the link shape.

kunal-mohta: @LaCuneta Okay. I will try this and get back to you.

LaCuneta:
This is looking pretty good to me.  I'd like to see it squashed down to a single commit and with a more descriptive commit message about what the change is.

PR #379, opened 2018-02-28
Refactor/Bugfix/Seismic: Restructure CoffeeScript sources (and more)

TheBizzle:
Sorry, I know this is an enormous blob of changes all rolled into one commit, and I definitely wouldn't like that if anyone else did it.  But I was reorganizing the folder structure and kept seeing other things that had been problematic for too long, and I went and fixed those, too.  One thing led to another, and then I was left with this enormous ball of changes all mixed in with the folder structure changes.  My apologies.

## File structure

  * No longer pancake CoffeeScript filenames (e.g. `applesorangesbananas`) but, instead, dasherize them (e.g. `apples-oranges-bananas`)
  * Delete TortoiseJS readme and folder, tear down the 'agent`/`communication` folder structure, and replace it all with this hierarchy:
    * javascripts/
      * (Misc. standalone crap, some of which used to be nested deeper in the folder structure)
      * beak/
        * babybehaviorspace
        * session-lite
        * tortoise
        * widgets/ (at the base level, several files that each used to be a part of `widgets.coffee` or `setupinterfaceeditor.coffee`)
          * initialize-ui (**this is the new entry point into the interface tab code**, which coordinates the actions of all of the other files)
          * config-shims (building the config objects that get passed on to the Tortoise engine (e.g. for plotting, printing to the output area, etc.))
          * skeleton (initializes our Ractive instance and holds the template for it)
          * handle-widget-selection (Ractive listeners for widget selection)
          * handle-context-menu (Ractive listeners for the context menu)
          * widget-controller (the old `WidgetController` class from `widgets.coffee`, plus some utility functions that it uses)
          * set-up-widgets (code for filling out widgets when they're first put into the UI, and when `WidgetController` creates new ones)
          * event-traffic-control (catch-all for Mousetrap and Ractive listeners that I wasn't able to think of a good, logical grouping for)
          * draw/ (3 files of View widget and drawing stuff)
          * ractives/
            * (Our various Ractive components, especially for widgets and their edit forms)
            * subcomponents/
              * (Smaller little components with few-to-no dependencies, which aren't very useful by themselves)
  * Move some things into a new file: `global-noisy-things.coffee`

## Broad changes

  * Use dasherized names for all Ractive events (whereas, before, it was a mixture of camelcase and dasherization)
  * No longer allow widgets to define for themselves whether or not the model will recompile when the widget is deleted; let `WidgetController` hold that information
  * No longer send a widget's triggers with the updated values every time that an edit form is submitted.  Instead, since the triggers are always the same for that type of widget, the triggers live inside the widget, and are extracted when needed.
  * Provide many more type annotations and `return` calls at the ends of functions of return type `Unit`

## Finer-grained changes

  * Rename the `AgentStreamController` class to "ViewController" (since that's what it really is, and how it's referred to throughout the rest of the code)
  * In same places where we were doing things like `oninit: -> @observe('x', [...])`, use the cleaner `observe: { x: [...] }` (and the same for if some other `@on('x', -> [...])` listener was being declared inside of `oninit`)
  * Rename `RactiveEditorWidget` to "RactiveModelCodeComponent" to be clearer about the fact that it's not a widget, and that it's not some general purpose thing
  * When code in a CodeMirror pane is read-only, pass around that variable as named "isReadOnly" instead of as "readOnly"
  * For the edit forms for Button, Label, and Monitor widgets, eschew separate `label` and `input` tags, in place of using the `labeledInput` subcomponent (which must have just been an oversight on my part before)
  * Actually sanitize the Info tab Markdown with Google Caja (we were doing this before, but then not actually using the sanitized value that we generated)

### context-menu.coffee

  * Rename `backgroundOptions` to "defaultOptions" (because the old name really didn't make much sense)
  * Do recursion properly in `alreadyHasA`, rather than trying to call the function by its original name (i.e. "alreadyHasAnOutput")
  * Properly lock the widget selection when the context menu is brought up for a widget (which is to say that a widget will now display as selected when right-clicked)

### session-lite.coffee

  * Shuffle around some code
  * Rename `exportnlogo` to "exportNlogo", because, seriously, WTF?
  * Call `widgetController.redraw()` right after `world.clearAll()` when recompiling code, because we need to read and run the updates from clearing the world before we clobber `Updater` with all of the stuff from the newly-compiled version of the model (how was this even working correctly before?)
  * Delete a bunch of dead code for calling the server to compile models, and delete the custom-defined AJAX function that only those functions were using.

### widget-controller.coffee

  * Where `WidgetController` used to take advantage of the fact that some values were declared out in the open and stored into `model` object, stop doing that.  `model` is no longer so easily accessible, so (**properly**) route those calls through `ractive.get`/`ractive.set`
  * Make the code for `updateWidget` (in all of its control-flow-heavy glory) a bit easier to read through
  * Instead of defining a `buttonEquals` and a `chooserEquals` and so on for each different type of widget, just, for each widget, provide a set of the properties that should be looked at for determining equality, and then generate the equality functions for all widgets in a single, unified place


TheBizzle:
The CI build seems to be failing due to being unable to resolve a WebJar (even though I didn't change the dependencies at all).  I assume that's a temporary outage of some sort.  Hopefully, you can verify the build locally.

LaCuneta:
Yeah, the CI failure is due to that `object.omit` Webjar SHA1 mismatch that has yet to be resolved.  Any "new to CI" branches will fail to build because of that, but any existing branches (like `staging`) actually continue to work it seems.  

No worries on the largeness of the PR and thanks for the detailed notes.  I'll try to get this reviewed today or tomorrow at the latest.

LaCuneta:
This looks very good to me, I think this helps to clarify the purpose of the different pieces which will hopefully make it even easier to work with. 

I was only able to find one issue after banging through a bunch of different models and looking through the code:  I think the file `test/assets/javascripts/TortoiseJS/TortoiseTest.coffee` needs to move to `test/assets/javascripts/beak/TortoiseTest.coffee` to keep the Mocha tests (meager as they are) happy.

TheBizzle:
Alright, I fixed this and the "widget area not resizing on widget deletion" bug.  You can see the patchfile for what I've pushed to this branch [here](https://gist.github.com/TheBizzle/b76fca8e0d6009adf503d473bfd64447).

PR #382, opened 2018-06-30
Integrate 'Jump to Procedure' Feature

diparthshah:
I am sorry for creating a new pull request, this happen due to moving project to my new system and mistake made while configuring the git on new system which caused CLA to break.  

I am currently keeping "Recompile Code" button at the centre in its original position and slightly moving procedure names dropdown away from left margin as you suggested. This changes are so made because we are thinking of using right portion of the "Recompile Code" button for another HTML element to be included for another feature of the same GSoC project. 

@TheBizzle


diparthshah:
After recent changes.

screenshot from 2018-07-02 00-37-28

PR #384, opened 2018-07-30
Integrate 'Code Auto-Complete' Feature

diparthshah:


app/assets/javascripts/beak/widgets/ractives/code-editor.coffee
```diff
@@ -6,6 +6,7 @@ window.RactiveModelCodeComponent = Ractive.extend({
   , lastCompiledCode:  undefined # String
   , lastCompileFailed:     false # Boolean
   , procedureNames:           {}
+  , autoCompleteStatus:  Boolean
```
LaCuneta: This can just be `false` to start, the `Boolean` object is the constructor for object booleans in Javascript.

diparthshah: Okay.

diparthshah: Will take few mins to push and squash

app/assets/javascripts/beak/widgets/ractives/code-editor.coffee
```diff
@@ -49,9 +50,48 @@ window.RactiveModelCodeComponent = Ractive.extend({
...
+      if word.slice(0, term.length) is term
+        word
+    )
+    if found.length > 0
+      return { list: found, from: from, to: to }
+
+  autoCompleteWords: ->
+    autoCompleteWords = Object.keys(@getProcedureNames())
+    keywords = new Set(window.keywords.all)
+    keywords = Array.from(keywords).map((item) -> item.replace("\\", ""))
```
TheBizzle: I think it makes for confusing code to redeclare variables, especially right after declaring them, especially when the two values are of different types (i.e. `Set` and `Array`).  Please rename one of the`keywords` or inline the `Set` value in the exactly one place where it is referred to.

app/assets/javascripts/beak/widgets/ractives/code-editor.coffee
```diff
@@ -74,6 +114,12 @@ window.RactiveModelCodeComponent = Ractive.extend({
                 on-click="recompile" {{# !isStale }}disabled{{/}} >Recompile Code</button>
           {{/}}
         </li>
+        <li class="netlogo-codetab-widget-listitem">
+          <input type='checkbox' class="netlogo-autocomplete-checkbox" checked='{{autoCompleteStatus}}'>
+          <label class="netlogo-autocomplete-label">
+            Auto Complete {{#if autoCompleteStatus}}Enabled{{else}}Disabled{{/if}}
```
TheBizzle: The standard way this is done throughout the rest of Galapagos is `Auto Complete {{# autoCompleteStatus}}Enabled{{else}}Disabled{{/}}` (that is, without the `if`s).  It's a small thing, but I'd prefer to see the `if`s removed.

app/assets/javascripts/beak/widgets/ractives/code-editor.coffee
```diff
@@ -49,9 +50,48 @@ window.RactiveModelCodeComponent = Ractive.extend({
...
+      from = to
+    found = options.words.filter( (word) ->
+      if word.slice(0, term.length) is term
+        word
+    )
+    if found.length > 0
+      return { list: found, from: from, to: to }
+
+  autoCompleteWords: ->
+    autoCompleteWords = Object.keys(@getProcedureNames())
```
TheBizzle: I think it's asking for trouble/confusion to give an inner variable (`autoCompleteWords`) the same variable name as the function that it is declared within.  Please rename one of them.

app/assets/javascripts/beak/widgets/ractives/code-editor.coffee
```diff
@@ -49,9 +50,48 @@ window.RactiveModelCodeComponent = Ractive.extend({
...
+    keywords = new Set(window.keywords.all)
+    keywords = Array.from(keywords).map((item) -> item.replace("\\", ""))
+    autoCompleteWords = autoCompleteWords.concat(keywords)
```
TheBizzle: This assignment is unnecessary.  Simply return `Object.keys(@getProcedureNames()).concat(keywords)` instead.

app/assets/javascripts/beak/widgets/ractives/code-editor.coffee
```diff
@@ -49,9 +50,48 @@ window.RactiveModelCodeComponent = Ractive.extend({
...
+    if token.string and /\S/.test(token.string[token.string.length - 1])
+      term = token.string
+      from = CodeMirror.Pos(cur.line, token.start)
+    else
+      term = ''
+      from = to
+    found = options.words.filter( (word) ->
+      if word.slice(0, term.length) is term
+        word
+    )
```
TheBizzle: This inner function does not strictly follow `filter`'s type signature.  `filter` is of type `((A) => Boolean) => Array[A]`.  So the function you give it should be of type `(A) => Boolean`.  But the function you're providing is of type `(A) => (A|Unit)` (because, when `if` succeeds, you return the word (`A`), and when it doesn't, you return nothing `null`/`Unit`).  This _works_, because JavaScript does weird things with treating `null` as `false` and treating "truthy" things (like `A`s) as `true`.  But doing this makes the code longer and more difficult to reason about.  This can be expressed more simply with `found = options.words.filter((word) -> word.slice(0, term.length) is term)`, which will be brief, clear (to me, at least), and rigorously typed.

app/assets/javascripts/beak/widgets/ractives/code-editor.coffee
```diff
@@ -6,6 +6,7 @@ window.RactiveModelCodeComponent = Ractive.extend({
   , lastCompiledCode:  undefined # String
   , lastCompileFailed:     false # Boolean
   , procedureNames:           {}
+  , autoCompleteStatus:    false
```
TheBizzle: Could you put type signature comments on these new variables (particularly `procedureNames`)?

diparthshah:
Hello @TheBizzle Thanks for the review, I'll make the changes to done.

app/assets/javascripts/beak/widgets/ractives/code-editor.coffee
```diff
@@ -5,7 +5,8 @@ window.RactiveModelCodeComponent = Ractive.extend({
   , isReadOnly:        undefined # Boolean
   , lastCompiledCode:  undefined # String
   , lastCompileFailed:     false # Boolean
-  , procedureNames:           {}
+  , procedureNames:           {} # HashMap
```
TheBizzle: The type should be `Object[T, U]`, but please fill in what the actual types of `T` (keys) and `U` (values) are.

diparthshah: `procedureNames:   Object[String, Number]`  ?

TheBizzle: `Object[String, Number]`

diparthshah: `procedureNames: Object[String, Number]` gives error.

TheBizzle: The type signature is a comment.  `procedureNames: {} # Object[String, Number]`

diparthshah:
Required changes are done.  @TheBizzle

PR #385, opened 2018-08-03
Integrate 'Code Usage' Feature

diparthshah:


app/assets/javascripts/beak/widgets/ractives/code-editor.coffee
```diff
@@ -78,16 +83,68 @@ window.RactiveModelCodeComponent = Ractive.extend({
...
       CodeMirror.registerHelper('hint', 'fromList', @netLogoHintHelper)
       @setupAutoComplete(@autoCompleteWords())
+      @setupCodeUsagePopup()
       return
 
     'recompile': (_) ->
       @setupAutoComplete(@autoCompleteWords())
       return
+
+    'jumpToUsage': (context, usagePos) ->
```
LaCuneta: Sorry, should've spotted this before, but our preferred event naming style would be hyphen-separated terms, so `'jump-to-usage'` in this case.

diparthshah:
@LaCuneta  Changes done.

PR #395, opened 2021-03-02
Fix wrong path of CodeMirror assets

daviddostal:
While I was working on another issue, I found a curious bug.

Some paths to CodeMirror assets (mainly on the `/differences` page) were wrongly pointing to `/assets/lib/codemirror`, but the files are currently located only in `/assets/codemirror`. As a result, CodeMirror didn't load on the `/differences` page.

The interesting thing is, this apparently didn't cause a problem on the [production website](https://www.netlogoweb.org/docs/differences), because there were still some old files (last changed in 2016) in `/assets/lib/codemirror` on the server. So the library could load both from the old/wrong and the current/correct location.

Correct me, if I'm missing something. I checked and the issue is present even on a local prod build of a fresh clone of the `production` (and master of course too) branch of [NetLogo/Galapagos](https://github.com/NetLogo/Galapagos).

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -62,7 +62,7 @@ class Application @Inject() ( assets: Assets
     def linkTag(x: String)   = s"""<link rel="stylesheet" media="screen" href="$x">"""
     def scriptTag(x: String) = s"""<script src="$x"></script>"""
 
-    val cssURLs = Seq( "lib/codemirror/lib/codemirror"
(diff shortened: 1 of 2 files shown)


LaCuneta (approved):
Nice catch!  Getting old files properly cleaned from the server on deployment is a step I've been meaning to look into for a while, but in this case I guess it was helping us out to leave the old ones behind.

Thanks for the fix.

PR #396, opened 2021-05-05
Bugfix: Fix some bugs caused by error-notification refactor

daviddostal:
Hello, I found (and fixed) some smaller bugs caused by the refactor of the error notification system, here is a summary:

**1. Cannot delete newly created widget:**

   When adding a new widget to a model and then deleting it, it would not get deleted, because of an exception:
   `widget-controller.js#removeWidgetById: this.ractive.get(...)[id] is undefined`.

   This error was fixed by passing the missing new parameter `reportError` to `setupWidget` in `widget-controller.coffee#createWidget`.

**2. Exception on duplicate widget vars**

   When multiple different widgets use the same variable, an error should be shown to the user, but now it throws an exception because of a missing function `showErrors` used in `event-traffic-control.coffee#rejectDupe`. (`showErrors` was replaced by the new error notification system.)

   To fix this, I used the new notification system instead of `showErrors`. I hope I used the new error notification system correctly, if there is a better way to do it, just let me know.

**3. Unneeded references to `showErrors`**

   While fixing the 2. issue, I also found one other reference to the now nonexistent `showErrors`, which is called from the `show-errors` event in `event-traffic-control.coffee`. The `show-errors` event is used in an on-click handler in `monitor.coffee` and `slider.coffee`.

   But `show-errors` was never called from the `on-click` handlers anyway. (To be fired from the `on-click` handlers, the event in `event-traffic-control.coffee` would need to be `'*.show-errors'` (with the wildcard star)).

   I don't think the click handlers are necessary for error reporting anymore, so I removed them and also the `show-errors` event, which should now be unnecessary. Do the on-click handlers referencing `show-errors` still have a purpose? I'm happy to edit the commit if this is the case.

### DIFF
--- a/app/assets/javascripts/beak/widgets/event-traffic-control.coffee
+++ b/app/assets/javascripts/beak/widgets/event-traffic-control.coffee
@@ -147,6 +147,10 @@ window.controlEventTraffic = (controller) ->
     )
     return
 
+  showWidgetErrors = (widget) ->
+    if not widget.compilation.success
+      controller.reportError('compiler', widget.type, widget.compilation.messages)
+
   # (String, String, Any) => Boolean
   renameGlobal = (oldName, newName, value) ->
 
@@ -225,6 +229,7 @@ window.controlEventTraffic = (controller) ->
   ractive.on('check-action-keys'        , (_, event)         -> checkActionKeys(event))
   ractive.on('create-widget'            , (_, type, x, y)    -> createWidget(type, x, y))
   ractive.on('drop-overlay'             , (_, event)         -> dropOverlay())
+  ractive.on('*.show-widget-errors'     , (_, widget)        -> showWidgetErrors(widget))
   ractive.on('track-focus'              , (_, node)          -> trackFocus(node))
   ractive.on('*.refresh-chooser'        , (_, nada, chooser) -> refreshChooser(chooser))
   ractive.on('*.reject-duplicate-var'   , (_, varName)       -> rejectDupe(varName))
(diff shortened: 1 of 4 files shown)


LaCuneta (changes requested):
Hey thank you so much for cleaning up the mess I left!

I confirmed your first two failure scenarios and the fixes for those look perfect to me.

I agree with removing the `show-errors` event, too.  There is one small change there, though.  That event was supposed to allow monitor and slider widgets to pop-up some info about why they were failing when they were clicked (the button widget does this itself).  So I think what would be better there would be to raise the new `nlw-compiler-error` event with the appropriate error messages when those widgets are clicked.  Example for `monitor.coffee`:

```
        <label class="netlogo-label {{errorClass}}" on-click="['nlw-compiler-error', 'monitor', widget.compilation.messages]">{{widget.display || widget.source}}</label>
```

To test that out you can make a NetLogo model with invalid code in the monitor (`count turtels`), then load it into NetLogo Web.  When you click the red text on the monitor it should alert to the issues with the monitor code.  Same change should work with the slider widget. 

Let me know if you'd prefer not to make that change and I'll merge this and handle it separately.  Thank you again!

daviddostal:
Thanks for the elaboration on reporting errors for `slider` and `monitor` widgets, that's a quite useful feature I didn't know about!

I decided to leave the widget error-reporting centralized and not move the code into the widgets, because there is a bit of logic required errors are shown on click only if the compilation failed. And either duplicating the code across widgets or making the `on-click` handlers fat would feel a bit dirty.

So there is a new `showWidgetErrors` function in `event-traffic-control.coffee`, which reports errors to the users, and I renamed the `show-errors` event to `show-widget-errors` to make its purpose a bit more clear.

Let me know if there is anything else you'd like to change!

LaCuneta (approved):
Re-tested, looks good to me.  Thank you again!

daviddostal:
> 
> 
> Re-tested, looks good to me. Thank you again!

Wow, that was fast! Glad to help!

PR #406, opened 2022-05-11
V2: Establish connection + functionality from NLW to HNW code (with refactored components and iframes)

cs2027:


TheBizzle:
The changes from this PR were remolded and rolled into #416.

PR #407, opened 2022-05-31
PR Bonanza 1/4 - Misc. Fixes

TheBizzle:
Alright!  I'm ready to start getting my branch merged into `master`.  I don't want to obliterate my hapless reviewer with a giant meteor slam of changes, however, so I'm planning on doing it in four parts:

  1. Misc. fixes
  2. Plot authoring
  3. HubNet Web
  4. HubNet Web authoring

(1) and (2) are already basically ready (caveat: there are a couple of small issues that need to be cleared up before (1) is merged, which I'll describe below.)

As for (3), it is probably mergeable, but I want to think a bit on some of the major refactoring that I recently did for it, and see if maybe I can't/shouldn't do more.

And (4)... I might hold off on for a bit, since the authoring tool's level of polish isn't nearly as high as the other things.

So what's in this PR?  Really, just a collection of miscellaneous fixes that I've accumulated while working on HubNet Web.  Most of the changes are pretty uncontroversial, but *some* of them might be otherwise.  We can discuss things, as needed.

However, before approving this for a merge, there are actually two thing that I need to resolve with @LaCuneta, which are causing me issues.  The first is the addition of the `!important` CSS keyword [here](https://github.com/NetLogo/Galapagos/commit/28b6df764a7bbd3587194debe1e8b56d1bcfdd78#diff-6060d8f357bfdf814bb3b6a3c693c394401ff589934ced056f3be04b89bbe0a7R20-R23).  And the second is addition of an `align-items` property [here](https://github.com/NetLogo/Galapagos/commit/33a31911d99148ab99241cc783752106710ba176#diff-3615135d6ed9d47ed9e7ac9da2ee4970e59c85c84fafc92316a4cb5d4d7c7e7dR191).

I expect that I could find ways to work around these changes, but they are changes that are dissonant with my original intentions when writing the code in question, so I think we need to discuss those things and reach an agreement about how to proceed.


TheBizzle:
> 1. With the `startLoop` change for the `thunk`: I'm just curious, did you see an instance where the workspace wasn't complete when the startup call happened?

I did, yes.  It was a while back, so I'm hazy on the exact circumstances, and I admit that they might not even be reproducible in the current state of HubNet Web, and it might have even required the very slow and messed up computer that I am no longer developing on, but it was causing me issues that every time I was doing... _something_ with a model with a `startup` procedure, I was getting an error from trying to run `startup` too soon.

If you want to drop this change, and wait until the day/month/year when I (or someone) can provide a reliable reproduction of the bug, I'm fine with that.

> 2. [How do we check model compilation statuses?]

There are some existing SBT commands that are relevant to this.  First, I run `sbt start`, which... previously launched the application in production mode.  When you do this, you get a message saying that this command is deprecated, and that you should use `sbt stage` and find find your executable in `target/universal/stage/bin`.  When you do *that*, you're told that `bin` is a directory, not an executable.  Which makes sense.  So then you run `target/universal/stage/bin/galapagos`, and it gives you a cryptic Java error.  And then you run `sudo target/universal/stage/bin/galapagos`, and it starts the server in production mode.

See?  Easy!

Is there an alternative that you prefer to this?  I'm not certain there's an easy way to launch Play for production without `stage`/`start`.  I *think* they treat it as a whole different beast, but I might be wrong.

LaCuneta:
> If you want to drop this change, and wait until the day/month/year when I (or someone) can provide a reliable reproduction of the bug, I'm fine with that.

I inspected the code at runtime there, and it's all a bit fuzzy for me; I can't imagine how it would happen.  But I could see how involving an additional outside component, like HubNet Web could complicate things, so I also don't mind adding this "just in case".  I suppose my primary concern is Future Devs might look at it and conclude that we knew for sure it was required.

I did notice while inspecting that that it could use some `return`s for CoffeeScript (one was already missing, I guess):

```coffeescript
    thunk =
      =>
        if workspace?
          if ProcedurePrims.hasCommand('startup')
            f = () -> ProcedurePrims.callCommand('startup')
            runWithErrorHandling('startup', @widgetController.reportError, f)
        else
          setTimeout(thunk, 5)
        return

    thunk()
    return
```

> See? Easy!

💯 

> Is there an alternative that you prefer to this? I'm not certain there's an easy way to launch Play for production without `stage`/`start`. I _think_ they treat it as a whole different beast, but I might be wrong.

I was going say "well, we could just write a test for it, right?" but it looks like it [already exists](https://github.com/NetLogo/Galapagos/blob/master/test/models/ModelCollectionCompilerTest.scala), so feel free to disregard.

TheBizzle:
Good call on the missing `return` statements.

I looked more deeply into the `startup` issue.  I *was* able to reproduce the problem, but, also, I concluded that my solution really wasn't appropriate for what the root issue was.  The proper fix is a small conditional check that is specific to HubNet Web.  As such, I've removed the commit with the `thunk` change from this PR.

TheBizzle:
Okay, I think all outstanding issues are now resolved.  The CI is failing, but I don't fully understand why.  I believe the first build of this branch worked.  But now it's giving some error, when booting SBT, along the lines of "Requires JDK 1.6+, but got JDK 11.0".  Was there some recent change made to the default JDK on Jenkins, or...?  I'm unclear on where the error is coming from.

@LaCuneta, note that I made the `!important` and `flex-row` changes that we talked about, which may cause some breakage in NetTango.

LaCuneta:
Thanks for the updates.

> The CI is failing, but I don't fully understand why. I believe the first build of this branch worked. But now it's giving some error, when booting SBT, along the lines of "Requires JDK 1.6+, but got JDK 11.0". Was there some recent change made to the default JDK on Jenkins, or...? I'm unclear on where the error is coming from.

You'll [need to rebase one more time](https://github.com/NetLogo/Galapagos/commit/4ab65ba5833ae65144f4d0b6df17ed42bfa823f4) as we had to update the JDK on Jenkins to 11.  I'll review once tests pass.

TheBizzle:
Jenkins seems happy now. :sunglasses: :+1:

LaCuneta (approved):
LGTM!

PR #408, opened 2022-06-06
PR Bonanza 2/4 - Plot Authoring

TheBizzle:


LaCuneta (changes requested):
Nice.  I ran it through some tests and played around and it worked well for editing plots and making/editing pens.  Most testing done in Edge/Chrome, but a little in Firefox, too.

I'm a little confused on why the `ColorModel` dependency got dropped.  It would seem to me Beak is always going to depend on Tortoise, so it's okay to pull that in and avoid duplicating the code.  What am I missing there?

## Issues

Using the right-click menu to **Create Plot** failed for me:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'nodeName')
  at a.getArgs (highcharts.js:363:342)
  at new a (highcharts.js:363:238)
  at a.chart (highcharts.js:363:284)
  at new HighchartsOps (highcharts.js:177:32)
  at set-up-widgets.js:89:12
  at setUpWidget (set-up-widgets.js:50:33)
  at WidgetController.createWidget (widget-controller.js:40:5)
  at createWidget (event-traffic-control.js:31:16)
  at Ractive.<anonymous> (event-traffic-control.js:291:12)
  at Ractive.caller (ractive.js:9114:40)
```

When expanding the **Plot setup commands** for a plot that has some setup commands, the text box showed up empty.  When I clicked on the text box the code appeared (Peppered Moths has some setup code for easy test/check).

There is no way to tell if I've used a duplicate pen name or plot name as I'm editing pens and plots.  If I add a pen to my plot with a duplicate name and set it up, then click **OK** I do get the error message about dup'd pen names, but when I go back to the plot I was editing I do not see the new pen I created at all to 'fix' it.  This is in contrast to variable widgets, where if I make a slider with the variable name already taken by a code `globals [...]` I get an error, but I can go back into the slider to fix my mistake.  I waffled on this being enhancement versus issue, but I think I lean towards issue as a user could lose code they wrote, which would be frustrating.  Since I couldn't make a new plot I couldn't test if created plots get "removed" too when duplicate, but obviously that would cause loss of work, too.  It'd be preferrable to get advance warning as soon as the duplicate name(s) are chosen, and then to get the chance to fix the mistake without wiping it out when hitting **OK** and getting the error message.  Or, I suppose, go the desktop route of not letting users close the window with **OK** at all while the plot or pen is in an error state.

## Minor Stuff / Maybe Enhancements

I'll leave these up to you to decide if they're worth doing pre-merge or not.

I think all code setup/update command areas should be vertically resizeable.  Maybe start a bit larger than they would if empty when they have many lines of code already set?  The Bacteria Hunt Speeds model has a longer **Plot update commands** section, as an example.

Click **Edit** on a plot, expand **Plot setup commands** (or update commands), two issues below.  These don't seem to occur with the pen commands.

- Only part of the text box displays the "text edit pipe/bar" mouse cursor, I'd expect the whole box to show.
- When I click anywhere in the text box, no cursor appears.  If I click inside the text box and start typing, words and the text caret appear.

I feel like clicking the header text of a collapsed area should expand/collapse that area (**Plot setup commands**, **Plot update commands**).  I put this in **Maybe** because, from experience with NetTango, I know users will request this about 10 minutes after publishing these changes.

## Definitely Enhancements

These definitely don't need to be done to merge.

I think pens should choose a unique color when added to a plot.

I think the first pen in a plot should default to "expanded" for fast/simple editing with single pen plots.

TheBizzle:
Thanks for the feedback and sanity checking!  The fact that "Create Plot" doesn't currently work is particularly egg-on-face....  It worked in a previous version of the branch!  I'm pretty certain that I broke it in the most-recent wave of refactoring that I did, cutting down on a 4x copy+paste of the plot-setup code. :disappointed: 

Regardless, I'll update the branch with a version that addresses most/all of these things, in the coming days.  In the meantime, I wanted to respond to the more rhetorical point:

> I'm a little confused on why the `ColorModel` dependency got dropped. It would seem to me Beak is always going to depend on Tortoise, so it's okay to pull that in and avoid duplicating the code. What am I missing there?

We can revert that, if you prefer.  I personally don't like Beak having explicit dependencies on Tortoise.  Theoretically, I would like Galapagos/Beak to be able to function without needing the Tortoise engine running in the same frame—maybe, in some sort of scenario where view updates are sent over the network, and the client should be lightweight and not require a dependency on a compiler or simulation engine simply to render those updates.  While that's the ideal for the HubNet Web client, it's mostly a minor and theoretical point.  This change brings us nowhere near full Tortoise-independence in Beak.

I'm of two minds about the code duplication.  On the one hand, I don't like it, which is why I previously batted around the idea of moving some parts of Tortoise into their own/separate repositories (e.g. `ColorModel`, `AgentModel`) as a solution to the "no Tortoise in Beak" problem.  We discussed this idea at some point, and I failed to convince you of the value of it, and I set the idea aside.

So then the options I see are to have a dislikeable dependency, or to duplicate the code.  Ultimately, though, the `ColorModel` code is simply a port of what's in desktop NetLogo, and (as far as I'm seeing) that code hasn't changed very much in the last decade, so I viewed this duplication as being relatively low-risk.

While I (obviously) find the argument that I just stated reasonable, I also think that your argument for keeping things how they were is reasonable, too.  Whichever.  It's easy for me to delete the commit if we don't want it.

TheBizzle:
Alright!  I've now implemented your feedback, and it should be ready for a re-review!

A few words about the changes:

  * I removed the commit that duplicated the `ColorModel` code and will do some investigation separately to see about refactoring the engine-adjacent code into a library that is separate from the engine itself.
  * You requested that the code containers be made vertically resizeable.  My way of addressing this was to drop the fixed height on the editors, and, instead, simply allow them to grow/shrink in height to have exactly enough lines for the current text therein (to a minimum of one line).
  * You requested that the first plot pen be expanded automatically for fast editing.  But this one didn't sit right with me.  I don't feel great about trying to guess what, exactly, the user's intentions were when they opened up the plot editor—for all I know, they would be annoyed by the open pen pane and then go out of their way to close it—so I've chosen to keep all of the pen panes closed by default.

LaCuneta (approved):
I re-tested and everything works well.  The pen and plot name validations work great.  Nice work!

PR #411, opened 2023-01-25
Events, queries, and auto-save

LaCuneta:
This PR adds a few features to NetLogo Web:

- An event system for model changes (authoring) and usage (widget interactions), that can be hooked into by users embedding models in other pages or by apps that are wrapping around NetLogo Web (like NetTango).
- A simple query system for getting information out of a model.  This is meant to help users embedding models as iframes who might want to extract data to use outside of NetLogo Web, perhaps in response to certain events.
- An auto-save feature built around the event system that allows resuming work in progress after page reloads and/or refreshes.  It's aware of the model source location, so it tracks progress on each URL, file upload, and New Model separately.  It also allows for quick reversion of changes made back to the original source.

### How to Test

For the event system, you can add the `debugEvents` query parameter to the URL of the `/launch` or `/web` pages to see the generated events logged to the web console.  If you're just reviewing the events, that should be enough.  For end-users embedding models as iframes by hot-linking or via static HTML exports, they also can use the `relayIframeEvents` query parameter on their model in order to get events automatically passed to their containing window for handling by code.  A full listing of events is contained in `listener-events.coffee`.

For the query system, you can add the `debugQueries` query parameter to the URL of the `/launch` page to attach a simple `makeQuery()` function to the `window`.  The query type strings for that method are listed in `debug-query-maker.coffee`.  The full query list is in `iframe-query-handler.coffee`.  In theory this query system doesn't just have to be used through an iframe.  Something like NetTango that embeds a NetLogo Web model could also use it, but at the moment I haven't made that switch.

The auto-save feature should be pretty self-explanatory (knock on wood).  I wound up not including a user prompt when they click to *Revert to Original*, because it wound up being pretty easy to add a *Undo Revert* button as well.  The ease of undoing the action if they didn't really want it made the prompt seem unnecessary.  I also did not add the *Set as Authoritative* button yet.  We can still look at that in the future.  I feel like it may be less useful if we have a real, granular undo/redo system in place.  But exporting a file or new model to an nlogo file will set the current authoritative source with the exported file name.

### Future Work

- Create documentation for the features, including tutorials for sending NetLogo Web event and model data off to remote storage for later review.  I'll be working on this over the next couple of weeks.
- Add the ability to raise user-defined events through the logging extension.
- Make sure NetTango is fully unified with the events system, and that all NetTango events are also exposed.
- Expand the auto-save feature into a true undo/redo system.  Ideally this would be a smarter system to track individual changes (like widget moves, plot renames, etc) so that they could be undone without a full session teardown and compile from nlogo source.
- Add the ability to hook into a user's remote storage, like GitHub, Dropbox, or Google Drive, in order to store work in progress there instead of just in the browser local cache.
- Add events for engine-specific things, where it makes sense.  At least prims like `reset-ticks`, `clear-all`, etc.  NetLogo desktop has events for turtle birth and death, which seems excessive to me, but I can see where having the option would be nice.
- Many, many more smaller items.


app/assets/javascripts/alert-display.coffee
```diff
@@ -155,16 +155,15 @@ class AlertDisplay
...
 
-  # (String, Exception) => Unit
-  reportRuntimeError: (source, exception, code) ->
+  # (CommonEventArgs, { messages: Array[String] }) => Unit
+  'extension-error': (_, {messages}) ->
+    @reportError(messages.join('<br/>'))
+    return
+
+  # (CommonEventArgs, { source: String, exception: Exception }) => Unit
+  'runtime-error': (_, {source, exception}) ->
```
TheBizzle, line 166: I assume that you have a very good reason for it, but why use dasherized names here, instead of camel-cased ones?

app/assets/javascripts/beak/rewriter.coffee
```diff
@@ -0,0 +1,24 @@
+# So far I haven't had a reason to turn this into a real class or to add the properties as real code somewhere.  But I
+# still wanted the basic shape of the Rewriter documented somewhere.  Note that due to changes in how NetTango Web
+# handles recompiles on block changes, some of these are no longer used by that app.  -Jeremy B January 2023
```
TheBizzle, line 3: Since every single member of this non-class is optional... how does one tell if they have a `Rewriter`?  It doesn't look like it can be done either nominally or structurally.  It would be nice if there were *some* way (be it by making a real class, or adding some `type` member to the class to identify its type, or even an `isRewriter` property that serves as a witness to a value's `Rewriter`-iness).

If the argument is that we shouldn't care whether or not something is a `Rewriter`, then:

  1. 😒
  2. This file kind of begs the question of why it exists

app/assets/javascripts/beak/tortoise.coffee
```diff
@@ -74,41 +71,92 @@ fromURL = (url, modelName, container, callback, rewriters = []) ->
...
+      )
+
       callback({
-        type:        'failure'
-      , source:      'compile-recoverable'
-      , session:     openSession(container, modelPath, name, compiler, rewriters, secondChanceResult, true)
-      , errors:      result.model.result
+        type:    'failure'
+      , source:  'compile-recoverable'
+      , session: session
```
TheBizzle, line 148: If you so desire, this line can simply be `, session`.

app/assets/javascripts/beak/widgets/ractives/status-popup.coffee
```diff
@@ -0,0 +1,56 @@
...
+      if @get('hasWorkInProgress')
+        @showStatus()
+
+    'close-status': () ->
+      @closeStatus()
+  }
+
+  showStatus: () ->
+    @set('isUp', true)
+    timer = 5
```
TheBizzle, line 20: I really think the pop-up disappears too quickly.  Even doubling its duration might not make it appear for long enough, but it really seems to go away faster than a person who was reading it for the first time would be able to read it.

app/assets/javascripts/beak/widgets/ractives/widget.coffee
```diff
@@ -190,32 +206,79 @@ RactiveWidget = RactiveDraggableAndContextable.extend({
...
+        on-drag="drag-widget"
+        on-dragend="stop-widget-drag">
+      </div>
+      {{/}}
+      """
+  }
+
+})
+
+RactiveValueWidget = RactiveWidget.extend({
```
TheBizzle, line 246: I don't know... it just feels like this should simply be in a separate file, and not be something that needs to be yanked out of the export object for `RactiveWidget`.

app/assets/javascripts/beak/widgets/widget-controller.coffee
```diff
@@ -271,24 +278,15 @@ defaultWidgetMixinFor = (widgetType, x, y, countByType) ->
...
-plotProps     = [ 'autoPlotOn', 'display', 'legendOn', 'pens', 'setupCode', 'updateCode', 'xAxis', 'xmax', 'xmin', 'yAxis', 'ymax', 'ymin']
-sliderProps   = ['default', 'direction', 'display', 'max', 'min', 'step', 'units', 'variable']
-switchProps   = ['display', 'on', 'variable']
-textBoxProps  = ['color', 'display', 'fontSize', 'transparent']
-viewProps     = ['dimensions', 'fontSize', 'frameRate', 'showTickCounter', 'tickCounterLabel', 'updateMode']
-# coffeelint: enable=max_line_length
+  propEq = (prop) ->
+    v1 = w1[prop]
+    v2 = w2[prop]
+    ((v1 is null or v1 is undefined) and (v2 is null or v2 is undefined)) or eq(v1)(v2)
```
TheBizzle, line 289: How does it happen that we have a widget with one of its properties un-defined?

app/assets/javascripts/beak/wip-listener.coffee
```diff
@@ -0,0 +1,142 @@
...
+    @session       = null
+    @reverted      = null
+
+  getWipKey: () ->
+    "#{@storagePrefix}#{@nlogoSource.getWipKey()}"
+
+  # (SessionLite) => Unit
+  setSession: (session) ->
+    @session = session
+    # This is necessary in the case where we didn't tell the session there was WIP at initialization becase we wanted to
```
TheBizzle, line 19: `s/becase/because`

app/assets/javascripts/beak/wip-listener.coffee
```diff
@@ -0,0 +1,142 @@
...
+    wipKey = @getWipKey()
+    @storage.set(wipKey, @reverted)
+    @reverted = null
+    return
+
+  # (String, String, String) => Unit
+  _storeWipInfo: (wipKey, newNlogo, title) ->
+    wipInfo = {
+      version:   WIP_INFO_FORMAT_VERSION
+      title:     title
```
TheBizzle, line 68: This is another instance of being able to simply write `title` instead of `title:     title`

app/assets/javascripts/beak/wip-listener.coffee
```diff
@@ -0,0 +1,142 @@
...
+
+    catch e
+      console.log("Unable to set work in progress, `getCurrentNlogo()` or `_setWip()` failed.", e)
+
+    return
+
+  # (String, String) => Unit
+  _updateForFileExport: (fileName, newNlogo) ->
+    source = new DiskSource(fileName, newNlogo)
+    # If we are current working on an imported file or a new document, the just-exported nlogo file has become our
```
TheBizzle, line 115: `s/current/currently`

app/assets/javascripts/notifications/debug-listener.coffee
```diff
@@ -0,0 +1,10 @@
+createDebugListener = (events) ->
+  debugListener = {}
+  events.forEach( (event) ->
+    debugListener[event.name] = (commonArgs, eventArgs) ->
+      console.log(event.name, commonArgs, eventArgs)
```
TheBizzle, line 5: You might prefer `console.debug`

app/assets/javascripts/notifications/listener-events.coffee
```diff
@@ -0,0 +1,331 @@
...
+      null
+
+# (Array[String | DependentArg], Array[Any]) => EventTypeArgs
+createNamedArgs = (argSettings, argValues) ->
+  namedArgs = {}
+  if (argSettings.length < argValues.length)
+    throw new Error("not enough arg settings for the values given")
+  argSettings.forEach( (argSetting, i) ->
+    argName = getArgName(argSetting, namedArgs)
+    if argName isnt null
```
TheBizzle, line 301: `if argName?` is also an option

app/assets/javascripts/notifications/listener-events.coffee
```diff
@@ -0,0 +1,331 @@
...
+  else
+    dependentValue = args[argSetting.sourceArg]
+    maybeCase = argSetting.cases.filter( (argCase) ->
+      argCase.sourceArgValues.includes(dependentValue)
+    )
+    if maybeCase.length is 1
+      foundCase = maybeCase[0]
+      foundCase.argToAdd
+    else
+      null
```
TheBizzle, line 292: The large majority of the upper `else` branch could be replaced with `argSetting.cases.find(f)?.argToAdd`.  This will, however, return `undefined` where it would previously have returned `null`, which will require an edit in `createNamedArgs` below.

app/assets/javascripts/pages/differences.js
```diff
@@ -1,8 +1,8 @@
-import "/codemirror-mode.js";
+import "/codemirror-mode.js"
```
TheBizzle, line 1: This is a `.js` file (not `.coffee`), so those semicolons *are* valid.  And the same applies in the other `.js` files that you changed in this folder.  I assume that you know this, because the change is repeated and looks intentional.

Personally, my policy has long been to avoid JS's semicolon inference whenever possible.  To provide a mild argument as to why: In my experience, without semicolons, syntactic errors can produce some weird (read: confusing) error messages.

🤷‍♂️

app/assets/javascripts/pages/simulation.js
```diff
@@ -1,39 +1,94 @@
...
+    params.set('url', maybeUrl)
+    window.location.search = params.toString()
+  }
+}
+
+const listeners = []
+
+const disableWorkInProgress = params.has('disableWorkInProgress')
+var wipListener       = null
+var getWorkInProgress = null
```
TheBizzle, line 34: One option here is change this to:

```javascript
const [wipListener, getWorkInProgress] = (() => {
  ...
  return [wl, gwip];
})();
```

Another option is to *not*, and just leave things how they are.  But I'm just speaking for Camp No-`var`s (Especially in Top-Level Scopes).

app/assets/javascripts/queries/debug-query-maker.coffee
```diff
@@ -0,0 +1,43 @@
...
+    , queries:    [query]
+    , sourceInfo: sourceInfo
+    }, '*')
+
+    sourceInfo
+
+listenForQueryResponses = () ->
+  window.addEventListener('message', (event) ->
+    if (event.data.type is 'nlw-query-response')
+      console.log(event.data)
```
TheBizzle, line 38: Again, you might prefer `console.debug`

TheBizzle (changes requested):
Ayy, a really useful change that required a whole lot of grungy work of mucking around in the guts of the NLW UI! 🎉  Nice, and good work! 😎👍  And it was executed well enough that you didn't even give me a lot to complain about.

I guess that the big, outstanding question that I have relates to the performance impact of these changes.  Up till now, obviously, people have been using NLW without this functionality.  I've gotta imagine that many—probably *most*—possibly even *the vast majority of*—page visits on netlogoweb.org won't use this functionality at all, but they'll still be generating and processing some logging events.

Do we have any sense of the performance impact of this?  Is there some easy way that we could make it opt-in/opt-out, such that events aren't sent if, for example, a certain flag in the main Ractive object is set to `false`?

app/assets/javascripts/beak/rewriter.coffee
LaCuneta: Sure, it can be an empty class for overriding for clarity.

app/assets/javascripts/alert-display.coffee
LaCuneta: It's because they're events, essentially, and our event name convention has been kebab case.

app/assets/javascripts/beak/widgets/ractives/status-popup.coffee
LaCuneta: :+1: I'll bump it up to 20 seconds.

app/assets/javascripts/beak/widgets/ractives/widget.coffee
LaCuneta: Sure.

app/assets/javascripts/pages/differences.js
LaCuneta: It was totally intentional.  I don't like dealing with the semicolons and haven't run into those weird, confusing errors you mention when not using them.  I'll to throw them back in.

LaCuneta:
For performance, that's definitely a big consideration.  I took this into account in two ways:

1. When the session is created, [it only creates a `Ractive.on()` for each event that is actually in the listeners provided](https://github.com/NetLogo/Galapagos/blob/listen-and-save/app/assets/javascripts/beak/session-lite.coffee#L69-L80).  That means for most events that occur, they'll fire a simple Ractive event that will have no listener and it'll quickly be done.  Event args are not created for these events that aren't listened for.
2. I tried to be minimal with the data gathered and passed with each event.  For instance, when widgets are changed or used we don't pass every property of the widget.  This means more work for the listener if they want more details, and that's what the query system is meant to provide.

I think there can be more mitigation of the ghost Ractive event firing by wrapping that up into something each eventing component can use.  That could be the simple boolean flag you suggest, or maybe an empty noop method call, but either way it would avoid the overhead of the Ractive event firing and hunting for a non-existent listener.

At the moment, though, I'm not too worried about this yet as all the events that exist so far are lower-frequency.  The ones that would generate the most noise are things like slider changes or button presses.  Subjectively I cannot tell a different just trying out a bunch of both (with logging disabled) on my phone on the experiments site and in production.  If you've got other ideas on how to reliably, repeatably benchmark the UI I'm definitely open to getting some hard numbers.

If you're okay with it, I'd save this refactor/optimization until I add the engine events.  Those seem much more likely to happen in large quantities (turtle hatch/death and tick alert) that could really slow down a model even when events aren't being used.

app/assets/javascripts/beak/widgets/widget-controller.coffee
LaCuneta: Oof, I should've written this down.  I want to say it was something like when changing the type of an input widget during an edit?  I don't remember the scenario 100%, but I do remember it seemed like a normal legit bug to me when I found it.

PR #415, opened 2023-08-25
Settings Work in Progress Feature

alizenart:
Adding a work-in-progress feature on the setting tab in NetLogo Web to see which models were recently modified and allow quick access via URL. The details include the Model Title, the URL to get to the model, and the Data Last Modified, displaying the most recently accessed model first.


app/assets/javascripts/settings-controller.coffee
```diff
@@ -59,14 +60,40 @@ template = """
...
+formatLinks = (wipStorage) ->
+  result = []
+
+  for url, model of wipStorage.inProgress
+    if model.title and model.timeStamp
+      formattedUrl = "/launch#http://" + url.replace(/^url:\/\//, '')
+      result.unshift({
+        modelTitle: model.title,
+        url: formattedUrl,
+        dataAccessed: new Date(model.timeStamp).toLocaleString()
```
LaCuneta, line 87: Seeing how the "Date Last Modified: " looks in the interface, I think maybe changing it to something like, "changed on" would be better.  Or maybe dropping it altogether, as it'll be clear from context what is being presented?  Also, it'd be great if the date string could be formatted like, **August 28th, 2023** just to make it a little more reader-friendly.

app/assets/javascripts/settings-controller.coffee
```diff
@@ -59,14 +60,40 @@ template = """
...
+  {{/each}}
+</ul>
+
 """
+formatLinks = (wipStorage) ->
+  result = []
+
+  for url, model of wipStorage.inProgress
+    if model.title and model.timeStamp
+      formattedUrl = "/launch#http://" + url.replace(/^url:\/\//, '')
```
LaCuneta, line 83: This still assumes the `http://` protocol, but if we're viewing a secure page it needs to be `https://` instead.

LaCuneta (changes requested):
Thanks for submitting this!  There are some smaller issues I left comments for.

In running this locally on my machine, I noticed three missed requirements I should've written down for you sooner.  Sorry about that.  Hopefully none of them are too difficult.

1. One of the data fields in the `inProgress` info for each WIP entry is `version`.  We should check that that version in local storage is the latest from `wip-data.coffee` and skip any that are older.  At the moment [the `WIP_INFO_FORMAT_VERSION` value is not exported](https://github.com/NetLogo/Galapagos/blob/fc542d725c8b0a54b6c156007939ac650f299656/app/assets/javascripts/beak/wip-data.coffee#L1), but it can be added to the exports in that file.  Another alternative is to update all the found entries to the latest version using the `WipData` class, but that doesn't seem like something the settings page should be doing automatically.

2. In addition to the `url://`-prefixed keys, there are also `disk://` and `new://` prefixes.  Theoretically it should be possible to re-open those as well, but at the moment there is no good way to do it.  Additionally we're not storing the full original NetLogo model in the local cache, only the modified version, so reverting wouldn't be possible with a disk or new-file model source.  As such, just skipping these will be fine for now.

3. One other feature the WIP system supports in NetLogo Web is storage tags.  These are primarily meant for users who are using models in classroom settings, but other users could use them to edit multiple copies of the same model without needing to rename them or fiddle with local files.  You can check how these work in [the authoring docs under **Saving Your Changes**](https://netlogoweb.org/docs/authoring).  The storage tag is prepended before the type in the key - so `mytag1:url://netlogoweb.org/libary/model.nlogo`.  The generated link for the list, then, would need to use that tag like `/launch?storageTag=mytag1#http://netlogoweb.org/library.model.nlogo` in order to get the proper WIP info when the model link loads.

Sorry again for the long this of changes.  Let me know if you have questions on anything, or feel free to message me on Slack about it.

alizenart:
Hi Jeremy,

Thank you for the feedback! I’ll take a look at it today and change it.

Sincerely,
Alison

app/assets/javascripts/beak/wip-data.coffee
```diff
@@ -72,4 +72,4 @@ class WipData
     @storage.set(wipKey, wipInfo)
     return
 
-export { WipData }
+export { WipData , WIP_INFO_FORMAT_VERSION}
```
LaCuneta, line 75: Nit: No space after `WipData`, prefer space before closing `}`

app/assets/javascripts/settings-controller.coffee
```diff
@@ -60,35 +62,57 @@ template = """
...
-    <li> <a href="{{url}}">{{modelTitle}}</a> <br>Date Last Modified: {{dataAccessed}}</li>
-    <br>
+    <li><a href="{{url}}">{{modelTitle}}</a> {{storageTag}} <div>{{dateAccessed}} at {{timeAccessed}}</div></li>
   {{/each}}
 </ul>
 
 """
+#Formatting Links for HTML display
 formatLinks = (wipStorage) ->
   result = []
```
LaCuneta, line 81: Nit: Since this variable holds multiple values for display, maybe pluralizing it `results` would be clearer?

public/stylesheets/settings.css
```diff
@@ -32,3 +32,12 @@
...
 }
+
+li {
+  margin-top: 15px;
+}
+
+.description {
+text-align: center;
+
+}
```
LaCuneta, line 43: Nit: Prefer blank line at end of file.

app/assets/javascripts/settings-controller.coffee
```diff
@@ -60,35 +62,57 @@ template = """
...
 
   for url, model of wipStorage.inProgress
-    if model.title and model.timeStamp
-      formattedUrl = "/launch#http://" + url.replace(/^url:\/\//, '')
+    if model.title and model.timeStamp and not(url.startsWith('disk') or url.startsWith('new')) and model.version == WIP_INFO_FORMAT_VERSION
+      storageTagOutput = ""
+      storageTag = url.split(':').shift();
+      if storageTag !="url"
```
LaCuneta, line 87: Nit: Prefer space after `!=` operator and string.

app/assets/javascripts/settings-controller.coffee
```diff
@@ -60,35 +62,57 @@ template = """
...
+        storageTagOutput = " (tag: " + storageTag + ")"
```
LaCuneta, line 88: Nit: Might as well use CoffeeScript's string interpolation syntax here, just to start getting in the habit: `" (tag: #{storageTag})"`

app/assets/javascripts/settings-controller.coffee
```diff
@@ -60,35 +62,57 @@ template = """
...
       result.unshift({
         modelTitle: model.title,
         url: formattedUrl,
-        dataAccessed: new Date(model.timeStamp).toLocaleString()
+        dateAccessed: new Date(model.timeStamp).toLocaleDateString('locale', {
+          year: 'numeric',
+          month: 'long',
+          day: 'numeric',
+        }),
+        timeAccessed: new Date(model.timeStamp).toLocaleTimeString('locale', {
```
LaCuneta, line 101: I tried `(new Date()).toLocaleString(undefined, { year: 'numeric', month: 'long', day: 'numeric', hour: '2-digit', minute: '2-digit', second: '2-digit' })` in the browser console and this gave me `'September 6, 2023 at 02:14:04 PM'` which seems like what we want.  Any reason not to combine these two separate bits into one?

app/assets/javascripts/settings-controller.coffee
```diff
@@ -60,35 +62,57 @@ template = """
...
+          second: '2-digit',
         })
+        storageTag: storageTagOutput
+
+      })
+
+  result.sort ((b, a) ->
+    dateTimeA = new Date("#{a.dateAccessed} #{a.timeAccessed}")
+    dateTimeB = new Date("#{b.dateAccessed} #{b.timeAccessed}")
+    dateTimeA - dateTimeB)
```
LaCuneta, line 113: This sort works, but it seems odd to take the formatted date/time strings and use the `Date()` constructor to parse them back into comparable values.  Can we instead just add the `timeStamp` to the results and use that here?

LaCuneta (changes requested):
Thanks for the updates, this is very close!  I left more comments to be addressed in the code.  It might seem like a lot, but I'm just trying to make sure we cover everything and keep it as clean as possible.  Feel free to ask here or on Slack if you have any questions.

alizenart:
Thank you for the feedback! I will edit these and then add another commit.

LaCuneta:
Thank you for the changes!  The last thing is the linting errors we see, most are about using the English operators instead of the character ones, `isnt` and `is` instead of `!=` and `==`, line lengths, and one is for a semicolon.  You can see those [here](https://netlogobuild.northwestern.edu/blue/organizations/jenkins/Galapagos/detail/PR-415/3/pipeline) or run `coffeelint` in an sbt session.

alizenart:
Thanks for the feedback! I will upload the new version right away.

PR #420, opened 2023-12-11
Adding locales languages options

alizenart:
Added Spanish, Japanese, and Portuguese into locales, changed the settings so the locales change accurately and adjustments (ie locale is Brazilian Portuguese, will set the language to Portuguese from Portugal until we add more languages)

Also updated the tortoiseVersion in the build.sbt


build.sbt
```diff
@@ -36,7 +36,7 @@ lazy val root = (project in file("."))
     TestAssets / JsEngineKeys.npmNodeModules := Nil
   )
 
-val tortoiseVersion = "1.0-88d5c3c"
+val tortoiseVersion = "1.0-2330ccf"
```
LaCuneta, line 39: This change shouldn't be necessary now, the latest Tortoise that the `master` branch is on has the latest I18N bundles.

build.sbt
```diff
@@ -71,7 +71,7 @@ Assets / unmanagedResourceDirectories += baseDirectory.value / "node_modules"
 // This is definitely not meant to be universal, just to get the info without the typical color codes -Jeremy B February
 // 2023
 def unAnsi(str: String): String =
-  str.replaceAll("\\u001B\\[\\d+m", "")
+  str.replaceAll("\\u001B\\\\d+m", "")
```
LaCuneta, line 74: Was this change intentional?  If so, what's it doing?

app/assets/javascripts/settings.coffee
```diff
@@ -13,7 +15,17 @@ clamp = (min, max, val) ->
...
-  locale: "en_us"
+  constructor: ->
+    @locale = navigator.language.replace('-', '_').toLowerCase()
+
+    exactMatch = locales.find((locale) => @locale is locale.code)
+
+    if exactMatch
+      @locale = exactMatch.code
+    else
+      matchingLocale = locales.find((locale) => @locale.startsWith(locale.code.split('_')[0]))
```
LaCuneta, line 26: Because we control the `locales` data, I suggest we pre-split out a new field for `languageCode` in that data instead of doing the split programmatically here.

app/assets/javascripts/settings-controller.coffee
```diff
@@ -138,4 +141,4 @@ createSettingsRactive = (container, storage, wipStorage) ->
   })
   ractive
 
-export { createSettingsRactive }
+export { createSettingsRactive , locales}
```
LaCuneta, line 144: Nit: Prefer space before closing `}`

app/assets/javascripts/settings.coffee
```diff
@@ -1,3 +1,5 @@
+import { createSettingsRactive, locales } from './settings-controller.js'
```
LaCuneta, line 1: `createSettingsRactive` is imported but not used.  I think it can be removed.

LaCuneta (changes requested):
Nice thank you for this.  It'll be good to get the locale actually set per the user's system!

I left a few comments in the code, let me know if you have questions on any of them.

app/assets/javascripts/settings.coffee
```diff
@@ -11,9 +13,26 @@ parseFloatOrElse = (str, def) ->
 clamp = (min, max, val) ->
   Math.max(min, Math.min(max, val))
 
+updatedLocales = locales.map (locale) ->
```
LaCuneta, line 16: So this is closer to what I meant, but I actually meant just hard-coding the `languageCode` values on the objects themselves in the `settings-controller.coffee` file.  So no CoffeeScript code should be required to do it programmatically.

app/assets/javascripts/settings.coffee
```diff
@@ -11,9 +13,26 @@ parseFloatOrElse = (str, def) ->
 clamp = (min, max, val) ->
   Math.max(min, Math.min(max, val))
 
+updatedLocales = locales.map((locale) ->
```
LaCuneta, line 16: With the hard-coded `languageCode` values in `locales`, I don't think this clone is necessary.  `updatedLocales` can be removed.

LaCuneta:
Closing as a new PR was opened

PR #421, opened 2024-02-01
Adding languages locale

alizenart:
I am adding the language locale, so that NetLogo can sense the browser's language, even if it is a different countries, so it can temporarily use the most fitting language.


LaCuneta (approved):
:+1: Nice work, thank you!

PR #422, opened 2024-02-10
Enhance deployment script with validation and user feedback

richardcmckinney:
Summary of Changes:

- Added checks for the presence of necessary command-line tools (`sbt`, `aws`, `cp`) before the execution of the script. This ensures that the script fails early with an informative message if a required tool is not available, guiding the user to resolve the issue before proceeding.
  
- Implemented input parameter validation to ensure that the script is called with exactly three arguments. This prevents the script from running with incomplete parameters, which could lead to unexpected behavior.
  
- Introduced more descriptive variable names (`S3_BUCKET_PATH`, `CLOUDFRONT_DISTRIBUTION_ID`, `ABSOLUTE_URL`) to improve code readability and maintainability.

- Included informational log messages to provide users with feedback about the script's execution process. These messages make the script more user-friendly by indicating the current action and any successful completions or errors that occur.

- Removed the final comment in the code, as the additional tests and checks now cover all aspects of the deployment process.

These enhancements make the script safer to use by preventing misconfigurations and providing clearer instructions to the user, both in terms of the script's requirements and its runtime operations.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -1,25 +1,45 @@
 #!/bin/bash
 
-# This script can be used to deploy Galapagos to a non-standard AWS location
-# AWS credentials will need to be configured for this script to work.  See the
-# AWS command line tools docs for how to do that.
+# Deploy Galapagos to a specific AWS S3 bucket and invalidate its CloudFront distribution.
 
-# Usage (from root of the Galapgos repo):
-# ./scripts/deploy.sh s3-target-bucket/target-folder DISTRIBUTION_ID url/path
+# Check for necessary command line tools and exit if not installed.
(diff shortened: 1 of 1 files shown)


LaCuneta (changes requested):
This is an interesting set of changes.  Just out of curiosity, were these changes done with a LLM or other AI coding tool?

I think the existing comments should be re-added as they give context on why the script would be used, as well as some useful examples of how to run it with our common sites.

The comments added don't seem helpful, they describe the commands almost verbatim, especially with the `echo` statements giving mostly the same text.

TheBizzle:
Closing for inactivity.

PR #427, opened 2024-11-07
Feature: Add a query parameter to override model title

kgmt0:
This will be useful in situations where it's not possible to control the filename of the model, which is currently used to set the model title. For example, Discourse automatically overrides the filename in the URL of uploaded attachments to the hash of the file.


TheBizzle:
Good stuff, thank you. 😎👍

PR #433, opened 2025-07-21
Update NetLogo Web Styling to (partially) match NetLogo Desktop V7.0.0-beta2

omargfh:
This PR brings NetLogo Web’s UI styling and widget behavior closer to NetLogo Desktop v7.0.0-beta2. Key additions and improvements:

### Highlights

* **Custom Slider Widget**
  Introduced a new Ractive`<customSlider>`to replace the native `<input type="range">` elements across:

  * Widget Sliders
  * Model Speed Slider (now with -/+ buttons and improved layout)

* **Switch Widget Cleanup**
  Updated the DOM and styling for switches to support better theming and consistency with new buttons. Now, it uses the same toggle as the one in NetLogo Desktop.

* **Styling Changes**
  Significant CSS cleanup across:

  * `netlogoweb.css` and `widgets.css` to match new theme colors and widget specs
  * Button states, disabled styles, input borders, slider tracks, and more

* **New Variables in `:root`**
  Unified color system and spacing constants for easier theme maintenance.

### Other Notes

* Fallback native inputs preserved (hidden) for screen readers and edge cases

### DIFF
--- a/app/assets/javascripts/beak/widgets/ractives/slider.coffee
+++ b/app/assets/javascripts/beak/widgets/ractives/slider.coffee
@@ -183,7 +183,7 @@ RactiveSlider = RactiveValueWidget.extend({
   }
 
   widgetType: "slider"
-  
+
   on: {
     'reset-if-invalid': (context) ->
       # input elements don't reject out-of-range hand-typed numbers so we have to do the dirty work


public/stylesheets/netlogoweb.css
```diff
@@ -1,32 +1,61 @@
+/*
+ * OpenSans
+ * */
+@import url('https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300..800;1,300..800&display=swap');
```
TheBizzle, line 4: Not legit.  NetLogo Web needs to bundle everything that it uses and to be usable without an internet connection.

app/assets/javascripts/beak/widgets/ractives/switch.coffee
```diff
@@ -108,6 +108,7 @@ RactiveSwitch = RactiveValueWidget.extend({
   minWidth:  35
   minHeight: 33
 
+
```
TheBizzle, line 111: Doesn't seem like a meaningful addition

app/assets/javascripts/beak/widgets/ractives/slider.coffee
```diff
@@ -170,10 +171,12 @@ RactiveSlider = RactiveValueWidget.extend({
   , breedVars:          undefined # Array[String]
   , errorClass:         undefined # String
   , internalValue:      0         # Number
+  , onValueChange: (val) ->
+      @set("internalValue", val)
+      @fire("widget-value-change")
   }
 
   widgetType: "slider"
-
```
TheBizzle, line 176: Seems unintended

TheBizzle (changes requested):
Pretty cool.  This is 90+% right, but there are some changes that I'd like to see.

- [x] You added a bunch of comments.  Most of them strike me as fairly self-evident things that don't need to be in the codebase.  If you *are* going to add a comment, however, it should be signed and dated, as we do throughout the NetLogo codebases.
- [x] After clicking the background of a Slider widget, the :arrow_left: and :arrow_right: keys should move slider knob in the corresponding direction
- [ ] "Edit" dialogs should give the option to use (or cease using) the old sizing
- [ ] When it comes to Switch widgets, these have a knob/track that is taller and narrower than it is in desktop NetLogo
- [ ] The "Edit" dialog for Note widgets has serious aesthetic issues, especially in Firefox
- [x] The "Opacity" slider added to Note widget "Edit" dialogs doesn't capture the full functionality.  Which would be addressed by the next point.
- [x] We should be using [the new color picker](github.com/NetLogo/ColorPicker), which was created precisely for the purpose of having the same color picker in both NLW and desktop NL.  It should be used in  should get used in: color Inputs, plot pen color picker, and the color pickers for Note background/foreground colors.
- [x] Monitor widget "Edit" dialogs should include the new "units" field, and support that functionality
- [x] The colored background should be brought back while in widget authoring mode; it's really not clear where you can and can't click to add widgets
- [x] When I make an Output widget and drag it around in the widget area, it's transparent compared to the other widgets, which makes it hard to see how my UI is actually going to look; the buttons behind the Output are easier to see than the Output that I'm trying to position.  Output, in particular, really needs to be more visible, since it's ~92% just a transparent white box right now, which makes it very hard to see.

@LaCuneta , your review is also requested here.  I mainly focused on functionality, so I'd especially appreciate you looking more closely at the technical details.

LaCuneta (changes requested):
Hey, this looks awesome!  I'm impressed how close you got the NL7 look-and-feel.  Some items below.  All of my testing was on Ubuntu 24 and latest Chromium.

The code items:

- [x] The `RactiveAsyncLoader` class doesn't appear to be used anywhere.
- [x] Is there a reason the `RactiveColorPicker` has to signal updates via callback functions (`onPick` and `onClose`)?  The more typical Ractive way to handle this would be to have the picker raise an event that the consumer (`RactiveIntColorInput`) could catch and handle.  The `RactiveEditFormDropdown` and it's `dropdown-changed` event would be a good example.
- [ ] As we discussed, I think the color picker code should be included via the npm registry as a normal package if possible.  I don't mind including it bundled for this PR as long as we get it updated in the near future.

The items found in my testing:

- [x] Many of the widgets have incorrect default sizes when adding them via authoring.  Vertical sliders and choosers aren't really usable after creation, it looks like.  Since they can't be expanded in the needed direction after creation, it can't be fixed inside NLW, either.
- [ ] Vertical sliders still have the slider drag left-to-right instead of bottom-to-top.  Heat Diffusion is a good model to test with.
- [x] The switch toggle doesn't change to a pointer finger on hover - it stays as the arrow.  The animation on the switch feels a little slow to me.
- [x] When I click on a slider it highlights the slide-bar with a border.  It's possible to click on that slide bar border (or the margin just past it) instead of the "main" bar, and it will not change the value even though it feels like it should to me.  It's easier to get that to happen clicken on the bottom side, at least for me.  This is very noticeable on the speed slider, where it's hard to actually hit the skinny bar.
- [x] The speed slider takes up a lot of vertical space now.  It probably already took up too much space before, though.  Maybe it can be a litter smaller and moved up to authoring-mode/commands-and-code-position line?
- [ ] If I have the Commands and Code set to be on the right side and I have a model that isn't too tall, in the color picker dialog the "OK" and "Cancel" buttons aren't visible . Related, in "Advanced" mode of the color picker there is a lot of empty space at the top and the bottom.  It appears like this is the space the full NetLogo color grid needs in "Simple" mode.

LaCuneta:
> It is used in `color-picker.coffee` but I imported it as `AsyncLoader` not as `RactiveAsyncLoader`. I will change the name.

Ah, apologies for goofing on my text search, but a rename would be appreciated.

> Ah, I'm not familiar with Ractive's way of doing things. React, WordPress, and vanilla all use callbacks. I'm assuming you define events in the Ractive's `on` field then you call `@findComponent(...).fire(...)` whenever you need need to use that event?

I think that's sort of backwards, but I might be reading it wrong.  To use the `RactiveEditFormDropdown` subcomponent as an example, when the `<select>` element changes, it detects that and does a `@fire('dropdown-changed', newValue)`.  Then where the dropdown is consumed, in this case it looks like in `HNWPenForm` in the `plot.coffee`, you add the `on: { "*.dropdown-changed": () -> }` handler to get that event and the data from it.  You shouldn't need to `@findComponent()` there for anything, as the `*.` part says "from this or any subcompnent".  

> Should we make those changes in this PR?

If possible, yes, I just don't want it to get lost and then in the future we'll start assuming it's there for a reason again.  Hopefully it's not too much work.

> @LaCuneta I agree. I'm not a fan of sandboxing and iFrames for use cases that are single components. If we publish it to NPM, can we change it to render to an object with a render target instead?

I'm totally good with that, it sounds like the same way the NetTango Web blocks component is brought in, and that works fine.

LaCuneta (approved):
I think this is good to go ahead and merge so others can check it out in staging.  I found the font bundling still isn't working (see below), but we can handle that separately.

These items I think need to be handled so we can bundle for the NL7 release:

- Make sure the "old size" bits are visualized correctly.  Thanks for adding the CSS class pass-through and edit boxes!  I'll keep working on this.
- In the exported HTML file for a model (using the **Export: HTML** button), the new font is not actually included.  When loading the HTML file I see the error loading it in the dev console.

These items can be taken care of as time allows, in my rough estimation of importance:

- See if we can reclaim some of the space the speed slider is taking up.
- See if we can avoid having to detect the type of color returned from the picker to parse.
- Get the color picker to live in the npm repository and be consumed as a normal dependency.

PR #437, opened 2025-08-23
Inlined Iframe when exporting and minor changes for 7.0.0 final changes

omargfh:
- Added traffic light controls to RactiveModal and integrated a close button for RactiveColorPicker.
- Enabled inline iframe dependencies for Galapagos export, ensuring color picker assets are embedded into HTML.
- Improved slider widget spacing to align with NL7 defaults.


LaCuneta (changes requested):
This looks alright to me.  I tested exported model HTML and the widget worked correctly as well as the color picker for label authoring.

- The new close button on the color picker modal doesn't work for me in latest Chromium on Ubuntu 24.  The dialog seems to close for a split second, but then re-appears.  The OK and Cancel buttons work fine.
- In the exported HTML, the `thin-strip-loader.gif` for the `RactiveAsyncLoader` gives a 404 error.  Would it be possible to just use some CSS effect or an inline SVG definition for this instead?  Just avoid needing to load an image entirely.  Something like what the `spinner.scala.html` does, but you shouldn't need a seperate view for it, just inline it into the async loader.

Ractive is now giving a couple warnings to the console it did not before:

- "Passing a `data` option with object and array properties to Ractive.extend() is discouraged, as mutating them is likely to cause bugs. Consider using a data function instead"  You should be able to add a `->` to the `data:` field of `RactiveModal` to clear this, making it generate the data object on create instead of re-using the same one.
- "<colorInput> Cannot use two-way binding on <input> element: hexColor is read-only. To suppress this warning use <input twoway='false'...>
printWarning @ ractive.js:590"  I think this should be a matter of adding that `twoway` setting to the `input` element in the template for `RactiveColorInput`.

Issue #373, opened 2017-11-08
Choosers with `true` and `1` options repeatedly flash on open in Chrome on macOS


LaCuneta:
The minimal case to reproduce is a model with a chooser with options `true` and `1`.  Opening the chooser will cause it to repeatedly flash until clicking somewhere else in the model, and the value cannot be updated.

In Firefox the window opens just fine, but both of the values aren't select-able.  Jason reports he doesn't see this problem in Opera or Chrome on Linux.  

After tracing through the code, the offending line is [this call to `@ractive.update()`](https://github.com/NetLogo/Galapagos/blob/master/app/assets/javascripts/TortoiseJS/agent/widgets.coffee#L332).  Removing that call removes the unwanted flashing behavior (or unable to select Firefox behavior).  The strange thing is the choosers are not being updated in the related code above the call that I can see, so there really isn't anything for Ractive to update with the choosers.  

This only affects models that mix `1`/`true` or `0`/`false` in chooser options, which should not be common.  The workaround for models that do need both in a chooser would be using alternative strings for the values and converting them in NetLogo code when changed.

LaCuneta:
Bumping the Ractive version to 0.9.9 (from 0.9.0) did not fix the issue.

TheBizzle:
I've also found that, on Linux, this happens in Firefox (but not Chrome) when the Chooser contains a list as an item.

PR #446, opened 2025-09-08
[Fix] Change RactiveChooser to reference options using their indices not values (Fixes #444)

### LINKED ISSUE #444: Chooser accepts duplicate or multiple congruent choices, causes a browser/OS level glitch
## Issue
Chooser accepts these as valid sets of choices 
- `[1, 1]`: Has the same element multiple times
- `[1, "1"]`: Has the same element value in `<option value=... />` multiple times**

> Clarification on the second case: What happens is that both choices get assigned the same value attribute on the `<option>` tag.

When multiple `<option>` present the same `value` attribute, and one of the duplicate choices is selected, clicking on the chooser to trigger the dropdown results in undefined behavior.

### MacOS 15.5, Chrome Canary on ARM (Browser-level Glitch)
https://github.com/user-attachments/assets/55260cb4-444d-463c-9f3e-1056760c962e

### MacOS 15.5, Firefox Developer on ARM (Input-level glitch)
https://github.com/user-attachments/assets/da9edb13-79f9-46b0-9963-1c020c8e5611

### Ubuntu 24.04.02 LTS, Chromium on ARM (Input-level glitch)
https://github.com/user-attachments/assets/40d4bb14-f34d-4758-b435-cf71f0495151

### Ubuntu 24.04.02 LTS, Firefox on ARM (Input-level glitch)
https://github.com/user-attachments/assets/9c72a202-36f9-4ce2-95d4-233bde4ce166

omargfh:
## Relevant Issue
Fixes #444

## Steps to Remedy Issue
1. Set the `value` attribute on `<option>` in `RactiveChooser` to use the `index` position of the choice rather than its value.
2. Bound the `<select>` to `widget.currentChoice` instead of the widget's `internalValue`.
3. Removed the `widget.currentValue` observer responsible for setting the `widget.currentChoice` (index) value on the widget controller.
4. Added a new local event `chooser-option-change` and set it to update the `internalValue` and `widget.currentChoice` based on the new selected index before firing `widget-value-change` to the `WidgetController`.

## Consequences and Side Effects
RactiveChooser refers to options using their indices not values, making each value unique even when congruent or equivalent to other values.**

> ** Note that if you have choices `[1, 2, 3, 1]` and have the _**last**_ `1` selected then append a new choice to the list, say, `[1, 2, 3, 4, 1]`, the event system will change your choice from the _**last**_ `1` to the _**first**_ `1` in the current implementation. This applies to options that match in _value_ and _type_. Congruent values (but different types) do not exhibit this behavior.

## Tested Environments

- [x] Chrome Canary on MacOS 15.5 (ARM)
- [x] Firefox on MacOS 15.5 (ARM)
- [ ] Chromium on Ubuntu 24 LTS (ARM)
- [ ] Firefox on Ubuntu 24 LTS (ARM)
- [ ] Chromium on Windows 10/11 (ARM)
- [ ] Firefox on Windows 10/11 (ARM)
- [ ] Chromium on Windows 10/11 (64-bit)
- [ ] Firefox on Windows 10/11 (64-bit)

### DIFF
--- a/app/assets/javascripts/beak/widgets/ractives/chooser.coffee
+++ b/app/assets/javascripts/beak/widgets/ractives/chooser.coffee
@@ -154,22 +154,54 @@ RactiveChooser = RactiveValueWidget.extend({
 
   widgetType: "chooser"
 
+  _setChoice: (idx) ->
+    widget          = @get('widget')
+    valid           = idx >= 0 and idx < widget.choices.length
+    # The compiler will complain if we set the currentChoice to an
+    # invalid index. However, we use that information internally
+    # to know whether the current value is valid or not and display
+    # the currentValue in the chooser when it is not valid. --Omar I. (09/23/2025)
+    compiler_idx    = if valid then idx else 0
+    choice          = widget.choices[idx] or widget.choices[0]
+
+    @set('internalValue', choice)
+    @set('widget.internalChoice', idx)
+    @set('widget.currentChoice',  compiler_idx)
+    choice
+
   on: {
+    'init': ->
+      widget = @get('widget')
+      @_setChoice(widget.currentChoice) if widget.currentChoice is -1
+        
     'chooser-option-change': (event) ->
-      widget       = @get('widget')
-      selectedIdx  = parseInt(event.node.value)
-      selectedChoice = widget.choices[selectedIdx] or widget.choices[0]
-      isSelectedIdxValid = selectedIdx >= 0 and selectedIdx < widget.choices.length
-      if isSelectedIdxValid
-        @set('internalValue', selectedChoice)
-        @set('widget.currentChoice', if selectedIdx >= 0 then selectedIdx else 0)
-        @fire('widget-value-change')
-        return
-      return false
+      @_setChoice(parseInt(event.node.value))
+      @fire('widget-value-change')
+  }
+
+  observe: {
+    'widget.currentValue': () ->
+      widget = @get('widget')
+      idx    = widget.choices.findIndex((c) -> c is widget.currentValue)
+      @_setChoice(idx)
+
+    'widget.choices': () ->
+      widget = @get('widget')
+      idx    = widget.choices.findIndex((c) -> c is widget.currentValue)
+      @set('widget.internalChoice', idx)
   }
 
   components: {
     editForm: ChooserEditForm
+  }
+
+  computed: {
+    selectedIndex: ->
+      widget = @get('widget')
+      if widget.internalChoice < 0
+        widget.choices.length
+      else
+        widget.currentChoice
   }
 
   eventTriggers: ->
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/beak/widgets/ractives/chooser.coffee
```diff
@@ -154,12 +154,18 @@ RactiveChooser = RactiveValueWidget.extend({
...
+    'chooser-option-change': (event) ->
+      widget       = @get('widget')
+      selectedIdx  = parseInt(event.node.value)
+      selectedChoice = widget.choices[selectedIdx] or widget.choices[0]
+      isSelectedIdxValid = selectedIdx >= 0 and selectedIdx < widget.choices.length
+      if isSelectedIdxValid
+        @set('internalValue', selectedChoice)
+        @set('widget.currentChoice', if selectedIdx >= 0 then selectedIdx else 0)
+        @fire('widget-value-change')
+        return
```
LaCuneta, line 167: Can either drop the `return` here or also `return false`.

app/assets/javascripts/beak/widgets/ractives/chooser.coffee
```diff
@@ -154,12 +154,18 @@ RactiveChooser = RactiveValueWidget.extend({
...
-    'widget.currentValue': () ->
-      widget        = @get('widget')
-      currentChoice = widget.choices.findIndex( (c) -> c is widget.currentValue )
-      @set('widget.currentChoice', if currentChoice >= 0 then currentChoice else 0)
-      return
+  on: {
+    'chooser-option-change': (event) ->
+      widget       = @get('widget')
+      selectedIdx  = parseInt(event.node.value)
+      selectedChoice = widget.choices[selectedIdx] or widget.choices[0]
```
LaCuneta, line 161: It won't cause an error, but it seems odd to me to get the `selectedChoice` before checking if the index is valid.  I think it could be moved inside the `if isSelectedIdxValid` conditional?  That might also remove the need to mix in the `or widget.choices[0]`.

app/assets/javascripts/beak/widgets/ractives/chooser.coffee
```diff
@@ -154,12 +154,18 @@ RactiveChooser = RactiveValueWidget.extend({
...
 
-  observe: {
-    'widget.currentValue': () ->
-      widget        = @get('widget')
-      currentChoice = widget.choices.findIndex( (c) -> c is widget.currentValue )
-      @set('widget.currentChoice', if currentChoice >= 0 then currentChoice else 0)
-      return
+  on: {
+    'chooser-option-change': (event) ->
+      widget       = @get('widget')
```
LaCuneta, line 159: I do prefer the assignments to be whitespace-aligned to the `=`, but it should be for all assignments in this block.  Having a standard 1 space before the `=` is fine, too, as long as it's consistent.

LaCuneta (changes requested):
Looking good, thank you!  Using the indices instead of the values looks like the right approach to me.  One issue I noticed is that with the loss of the `widget.currentValue` observer, the chooser no longer updates when I change its value through code instead of the UI.  If you load Wolf Sheep Predation it should default to a `model-version` with `"sheep-wolves-grass"` as the value.  If you run `set model-version "sheep-wolves"` in the Command Center, the widget will not update, but the backing global will.

I also test #373 with the new code, and it resolves that older issue as well.

I also left a few notes in the code.  Let me know if you have questions on anything.

LaCuneta (changes requested):
Thanks for the updates!  I left a few more code comments on small things.

The update to add in an unexpected value to the choices `<select>` is clever!  But this is a weird situation where the engine (Tortoise) should be rejecting that unexpected choice as an error (as NetLogo desktop does).  This is a weird situation as some widget global variables don't follow this "out of band error" rule (sliders).  For now, I think setting the chosen index to `-1` so the select shows blank is fine, until we get the error properly thrown in Tortoise sometime later on.

LaCuneta (approved):
Thanks for the updates!  I tested it out and things are working as I'd expect now.  I had a few comments in the code for review, but I'm just going to merge and make some updates so I can keep working towards a release.  Just let me know if you have questions on anything.

PR #449, opened 2025-10-14
[Fixes #448] Draft PR for VPAT Compliance

### LINKED ISSUE #448: Improve keyboard accessibility of NetLogo Web
@jzkelter tells me that

> VPAT ("Voluntary Product Accessibility Template") [is something that] many universities require (and probably public schools) before you can be a vendor for them. It includes this "All [functionality](https://www.w3.org/TR/WCAG20/#functiondef) of the content is operable through a [keyboard interface](https://www.w3.org/TR/WCAG20/#keybrd-interfacedef) without requiring specific timings for individual keystrokes..."

In this regard, we are deficient in a select few areas.

- [x] You can't tab to "forever" buttons (which just seems like a bug)
- [x] You can tab through most controls and interactable widgets, but not to the "Code" (etc.) panes
- [x] You should be able to escape text areas (e.g. the Code tab's editor) to resume tabbing through the page
  * My suggested solution: `Esc` jumps the focus up out of any editor
- [x] While not strictly necessary for this, it would nice to be able to quickly jump between the "Command Center", "Info", and "Code" panes
  * My suggested solution: In desktop NL, we use `Ctrl` or `Command` (i.e. `Ctrl+1`, `Ctrl+2`, `Ctrl+3`).  This _does_ pose a point of contention in the browser, since those are theoretically hotkeys used for navigating tabs.  But, honestly, I doubt that people use those shortcuts nearly as much as `Ctrl+Tab` and `Ctrl+Shift+Tab`.  We're probably fine overriding those shortcuts.
- [x] The current widget tabbing order can seem very random, because it's probably based off of whatever the order of the widgets is in the `.nlogox` file.  It might make sense to assign tab order based on position, left to right, top to bottom.  Maybe. 🤷‍♂️

omargfh:


app/assets/javascripts/beak/widgets/accessibility/key-combo.coffee
```diff
@@ -0,0 +1,40 @@
...
+    'up': "↑",
+    'down': "↓",
+    'left': "←",
+    'right': "→",
+    'escape': "Esc",
+  }
+
+  # Array[String]
+  getKeys: -> [
+    ...Object.entries(@modifiers).filter(([_, v]) => v).map(([mod, _]) =>
```
TheBizzle: I consider ```coffee
  ...some.long.multiline(
    computation
  ),
  if pred then x else y
]
```

to really lack in clarity and readability.  I'd really prefer you to store these complex values into variables, and then assemble the output array from the variables.  I don't usually value concision or cleverness in code.

File: app/assets/javascripts/beak/widgets/accessibility/key-combo.coffee [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+  # { String: String }
+  keyStringTable: {
+    'up': "↑",
+    'down': "↓",
+    'left': "←",
+    'right': "→",
+    'escape': "Esc",
+  }
+
+  # Array[String]
```
TheBizzle: This is `() => Array[String]`, not `Array[String]`

app/assets/javascripts/beak/widgets/accessibility/keybind.coffee
```diff
@@ -0,0 +1,64 @@
+import { isMac } from "./utils.js"
+import { KeyCombo } from "./key-combo.js"
+
+export class Keybind
+  # id: string
+  # cb: (ractive, KeyboardEvent, combo) => Boolean | Unit
+  # comboStrs: https://craig.is/killing/mice (excluding sequences)
```
TheBizzle: A URL is not a type

app/assets/javascripts/beak/widgets/accessibility/key-combo.coffee
```diff
@@ -0,0 +1,40 @@
...
+  # (String) => Unit
+  constructor: (comboStr) ->
+    @comboStr = comboStr
+    parts   = comboStr.toLowerCase().split('+').map((x) => x.trim())
+    @key    = parts.pop()
+    @modifiers = {
+      ctrl:  parts.includes('ctrl')  or parts.includes('control')
+      , alt:   parts.includes('alt')   or parts.includes('option')
+      , shift: parts.includes('shift')
+      , meta:  parts.includes('meta')  or parts.includes('cmd') or parts.includes('command') or parts.includes('mod')
```
TheBizzle: Odd-dentation of the `ctrl` line

app/assets/javascripts/beak/widgets/accessibility/keybind.coffee
```diff
@@ -0,0 +1,64 @@
+import { isMac } from "./utils.js"
+import { KeyCombo } from "./key-combo.js"
+
+export class Keybind
+  # id: string
```
TheBizzle: `string` here, but `String` down below (and throughout the rest of this codebase)

app/assets/javascripts/beak/widgets/accessibility/keybind.coffee
```diff
@@ -0,0 +1,64 @@
...
+  # description: String | undefined
+  # conditions: Array[((ractive) => Boolean)]
+  # keybinds (keybinds): Array[Keybind]
+  constructor: (name, description, conditions, keybinds) ->
+    @name = name
+    @description = description
+    @conditions = conditions
+    @keybinds = keybinds
+
+  # ractive => Boolean
```
TheBizzle: I'm not understanding the logic behind `Mousetrap` getting a capital letter at the beginning, but `ractive` having a lowercase letter at the beginning.  In this codebase, all types are considered to start with capital letters.  Lowercase first letters are reserved for variables/parameters.

app/assets/javascripts/beak/widgets/accessibility/utils.coffee
```diff
@@ -0,0 +1,45 @@
...
+  "link": ["a[href]", "area[href]"].map(unlessNoFocus),
+  "media": ["audio[controls]", "video[controls]", "iframe", "embed"].map(unlessNoFocus),
+  "contenteditable": ['[contenteditable]:not([contenteditable="false"])'].map(unlessNoFocus),
+  "role": [
+    '[role="button"]', '[role="link"]', '[role="textbox"]',
+    '[role="search"]', '[role="combobox"]', '[role="listbox"]',
+    '[role="menu"]', '[role="menubar"]'
+  ].map(unlessNoFocus),
+  "tabindex": ['[tabindex]'].map(unlessNoFocus)
+})
```
TheBizzle: You'd probably be better off writing `unlessNoFocus` to just be something that you call once to crawl this entire object.  Every item in every value here is getting the `noFocus` thing run on it.

app/assets/javascripts/beak/widgets/accessibility/utils.coffee
```diff
@@ -0,0 +1,45 @@
...
+
+sortByTabIndex = (a, b) ->
+  aTabIndex = parseInt(a.getAttribute('tabindex') or '0')
+  bTabIndex = parseInt(b.getAttribute('tabindex') or '0')
+  if aTabIndex < bTabIndex then -1
+  else if aTabIndex > bTabIndex then 1
+  else 0
```
TheBizzle: Conditional format:

```coffee
if pred1
  x
else if pred2
  y
else
  z
```

app/assets/javascripts/beak/widgets/accessibility/widgets.coffee
```diff
@@ -0,0 +1,31 @@
+# Rect: { x, y, width, height }
+# Rect -> Rect -> Int (comparison result)
+export comparePositionsRowMajor = (a, b) ->
+  { x: axcor, y: aycor, width: aw, height: ah } = a
+  { x: bxcor, y: bycor, width: bw, height: bh } = b
+  if aycor isnt bycor then aycor - bycor
+  else if axcor isnt bxcor then axcor - bxcor
+  else if aw isnt bw then aw - bw
+  else if ah isnt bh then ah - bh
+  else 0
```
TheBizzle: `if`/`else` formatting

app/assets/javascripts/beak/widgets/event-traffic-control.coffee
```diff
@@ -233,27 +241,63 @@ controlEventTraffic = (controller, performUpdate) ->
...
+    else
+      return
 
+    ractive.set(showPropertyName, showPropertyValue)
+
+    if showPropertyValue
+      componentToFocus = {
+        "Code": "codeEditor",
+        "Info": "infoeditor",
+        "Console": "console"
```
TheBizzle: Is this right?  One is `codeEditor` (capital `E`), while the other is `infoeditor` (lowercase `e`)?

app/assets/javascripts/beak/widgets/ractives/info.coffee
```diff
@@ -16,11 +18,15 @@ RactiveInfoTabEditor = Ractive.extend({
...
 
-    infoTabEditor.on('change', =>
-      @set('rawText', infoTabEditor.getValue())
-      @set('info',    infoTabEditor.getValue())
+    @infoTabEditor.on('change', =>
+      @set('rawText', @infoTabEditor.getValue())
+      @set('info',    @infoTabEditor.getValue())
     )
 
+  focus: ->
```
TheBizzle: Also missing type annotation

app/assets/javascripts/beak/widgets/ractives/keyboard-help.coffee
```diff
@@ -0,0 +1,140 @@
...
+
+RactiveKeyboardHelp = Ractive.extend({
+
+  data: -> {
+    id: "keyboard-help-dialog"
+    isVisible: false
+    isOverlayUp: false
+    wareaHeight: 600
+    wareaWidth: 600
+    keybinds: [] # see accessibility/keybinds.js
```
TheBizzle: See there for what?  Who wrote this?  When?

app/assets/javascripts/beak/widgets/ractives/keyboard-help.coffee
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  generateGroups: ->
+    groups = @get('keybinds')
+    return '' unless groups and groups.length > 0
```
TheBizzle: Early exit

`unless`

app/assets/javascripts/beak/widgets/ractives/keyboard-help.coffee
```diff
@@ -0,0 +1,140 @@
...
+      if key is "Escape"
+        @fire('close-help')
+        false
+  }
+
+  show: ->
+    @set('isVisible', true)
+
+  hide: ->
+    @set('isVisible', false)
```
TheBizzle: Recommend ending all functions that return `Unit` with `return`.

app/assets/javascripts/beak/widgets/ractives/tab.coffee
```diff
@@ -0,0 +1,78 @@
...
+      name:             undefined  # String
+    , title:            undefined  # String
+    , show:             false      # Boolean
+    , 'focus-target':   undefined  # String (Selector)
+    # https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView
+    , 'scroll-target':   undefined # String (Selector) defaults to focus-target
+    , 'scroll-behavior': 'auto'    # String ('auto' | 'smooth')
+    , 'scroll-block':    'nearest' # String ('start' | 'center' | 'end' | 'nearest')
+    , 'scroll-inline':   'nearest' # String ('start' | 'center' | 'end' | 'nearest')
+    , 'scroll-container': 'all'    # String ('all' | 'nearest')
```
TheBizzle: Why do it this way?  I'm not sure we have another place where we store variables dasherized in an object.  Might be a good reason for it, but it's not immediately obvious to me.

app/assets/javascripts/beak/widgets/ractives/tab.coffee
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  focus: ->
+    target = @find(@get('focus-target'))
+    if target?
+      options = @get('scroll-options') or {}
+      target.focus(options)
+    return
+
+  scrollIntoView: ->
```
TheBizzle: Missing type

app/assets/javascripts/beak/widgets/ractives/tab.coffee
```diff
@@ -0,0 +1,78 @@
...
+
+  on: {
+    keydown: ({ original: event}) ->
+      if event.key is " " or event.key is "Enter"
+        labelId = "tab-#{@get('name')}"
+        @set('show', not @get('show'))
+        event.preventDefault()
+        event.stopPropagation()
+        return false
+      return true
```
TheBizzle: `if`/`else` form

Needless `return`s

app/assets/javascripts/beak/widgets/ractives/tab.coffee
```diff
@@ -0,0 +1,78 @@
...
+
+  focus: ->
+    target = @find(@get('focus-target'))
+    if target?
+      options = @get('scroll-options') or {}
+      target.focus(options)
+    return
+
+  scrollIntoView: ->
+    name = @get('name') or 'unknown'
```
TheBizzle: There are a lot of places where you use `or` like this.  In CoffeeScript, I prefer to only see `or` used for logical comparisons.  `@get('name') ? 'unknown'` achieves what you want.

app/assets/javascripts/beak/widgets/accessibility/key-combo.coffee
```diff
@@ -0,0 +1,40 @@
+import { isMac } from "./utils.js"
+
+export class KeyCombo
+  # (String) => Unit
+  constructor: (comboStr) ->
+    @comboStr = comboStr
+    parts   = comboStr.toLowerCase().split('+').map((x) => x.trim())
+    @key    = parts.pop()
```
TheBizzle: This is actually something like

```coffee
[@key, parts...] = comboStr.toLowerCase().split('+').map((x) => x.trim()).reverse()
```

app/assets/javascripts/beak/widgets/accessibility/utils.coffee
```diff
@@ -0,0 +1,45 @@
...
+
+export getAllFocusableElements = (container) ->
+  selector = Object.values(accessibilitySelectors).flat().join(', ')
+  elements = Array.from(container.querySelectorAll(selector))
+                  .filter((el) -> el.offsetWidth > 0 or el.offsetHeight > 0)
+  elements.sort(sortByTabIndex)
+  return elements
+
+export offsetFocus = (container, element, offset) ->
```
TheBizzle: Missing types here and above

app/assets/javascripts/beak/widgets/event-traffic-control.coffee
```diff
@@ -233,27 +241,63 @@ controlEventTraffic = (controller, performUpdate) ->
...
-  mousetrap.bind(['ctrl+shift+h', 'command+shift+h']        ,           -> ractive.fire('hide-resizer'))
-  mousetrap.bind('del'                                      ,           -> ractive.fire('delete-selected'))
-  mousetrap.bind('escape'                                   ,           -> ractive.fire('deselect-widgets'))
+  setTab = (tabName, options = {}) ->
+    tabCanonicalName = tabName.charAt(0).toUpperCase() + tabName.slice(1).toLowerCase()
 
-  mousetrap.bind('?', onQMark)
+    validTabs = ['Console', 'Code', 'Info']
+    unless tabCanonicalName in validTabs
+      console.error("Invalid tab name: #{tabName}. Valid tabs are: #{validTabs.join(', ')}")
```
TheBizzle: IDK, is `console.error` really the right one?  I feel like, if it's truly an important error, we display it with `alert`, so they actually see it.  Otherwise, it's just a warning (and, thus, `console.warn`).

app/assets/javascripts/beak/widgets/ractives/button.coffee
```diff
@@ -297,9 +297,9 @@ RactiveButton = RactiveWidget.extend({
 
     standardButton:
       """
-      <button id="{{id}}" type="button" style="{{dims}}"
+      <button id="{{id}}" type="button" style="{{dims}}" {{attrs}}
               class="netlogo-widget netlogo-button netlogo-command{{#widget.oldSize}} old-size{{/}}{{# !isEnabled }} netlogo-disabled{{/}} {{errorClass}} {{classes}}"
-              on-click="@this.fire('activate-button')">
+              on-click="@this.fire('activate-button')" aria-disabled="{{# !isEnabled }}true{{else}}false{{/}}">
```
TheBizzle: Isn't this just `aria-disabled="{{isEnabled}}"`?  This patterned gets repeated a lot below.

app/assets/javascripts/beak/widgets/ractives/keyboard-help.coffee
```diff
@@ -0,0 +1,140 @@
...
+    return '' unless groups and groups.length > 0
+
+    groups = groups.map((g) =>
+      keybinds = (g.keybinds or []).map((kb) ->
+        description = kb.metadata?.description or 'No description available'
+        combos = (kb.combos or []).map((combo) -> { keys: combo.getKeys() })
+        { description: description, combos: combos }
+      )
+      disabled = not g.meetsConditions(@root)
+      { name: g.name, description: g.description, keybinds: keybinds, disabled }
```
TheBizzle: As you surely know, `{ keybinds: keybinds }` is just `{ keybinds }`

app/assets/javascripts/beak/widgets/ractives/subcomponent/code-container.coffee
```diff
@@ -12,6 +12,8 @@ RactiveCodeContainerBase = Ractive.extend({
   , injectedConfig: undefined # Object
   , onchange:       (->)      # (String) => Unit
   , style:          undefined # String
+  , tabindex:       undefined # String
+  , 'aria-label':   undefined # String
```
TheBizzle: I don't follow.  We define `aria-label` in the `data`, and then never read it?

app/assets/javascripts/beak/widgets/accessibility/key-combo.coffee
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+  # Print methods
+  # { String: String }
+  keyStringTable: {
+    'up': "↑",
+    'down': "↓",
+    'left': "←",
+    'right': "→",
+    'escape': "Esc",
```
TheBizzle: No need to explicitly put the keys into strings

app/assets/javascripts/beak/widgets/accessibility/key-combo.coffee
```diff
@@ -0,0 +1,40 @@
...
+    @key    = parts.pop()
+    @modifiers = {
+      ctrl:  parts.includes('ctrl')  or parts.includes('control')
+      , alt:   parts.includes('alt')   or parts.includes('option')
+      , shift: parts.includes('shift')
+      , meta:  parts.includes('meta')  or parts.includes('cmd') or parts.includes('command') or parts.includes('mod')
+    }
+
+  # Print methods
+  # { String: String }
```
TheBizzle: My first read of this is "`keyStringTable` is an object with only one key: `String`, which is of type `String`".  On the second or third parse, my brain understood that you were specifying that the keys are strings, but, generally, you don't have to do that, because JavaScript demands that the keys of all objects are strings, regardless.  Sometimes, I'll explicitly point out that I'm trying to only put numbers in as keys (for example), but a key type of `String` should be the first assumption, since it's the only true reality.  I'd just write this type as `Object[String]`.

TheBizzle (changes requested):
:sunglasses: 

Thanks for taking care of this! :bowing_man: 

I left a ton of comments here, but it's nearly all code style stuff.  I'm not sure if you did or didn't do this, but always make sure to go over the diffs of your changes before submitting, to check for errors and see if you can think of any ways you might have been able to do things better.

Regardless, the substance of the functionality seems pretty good.  I like that you were able to isolate most of the CoffeeScript changes to a separate `accessibility` package.  `keybinds.coffee` seems particularly nice.  I really like it when powerful things can just be simple and declarative.

There's no rush for addressing these things, I think, as I am _not_ trying to get it into the NLW bundle for 7.0.1.

app/assets/javascripts/beak/widgets/ractives/tab.coffee
omargfh: t goes something like this: these are DOM props, so they are kind of attributes, and attributes are kebab cased. This is totally a mental bias on my part so I can change it.

app/assets/javascripts/beak/widgets/accessibility/keybinds.coffee
```diff
@@ -0,0 +1,181 @@
+import { offsetFocus } from "./utils.js"
+import { isMac } from "./utils.js"
+import { Keybind, KeybindGroup } from "./keybind.js"
+
+modKey = if isMac then "command" else "ctrl"
+
+keybinds = [
+  new KeybindGroup(
+    "General Shortcuts",
+    undefined,
```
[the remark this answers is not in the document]
omargfh: `undefined` means _does not exist_ while `""` means _empty_. In the conditional check, it is natural to ask _Did the developed define the description?_ and if not, we don't render the description HTML.

app/assets/javascripts/beak/widgets/accessibility/utils.coffee
```diff
@@ -0,0 +1,45 @@
+noFocusSelector =  [
+  ':not([tabindex="-1"]):not([disabled]):not([aria-hidden="true"]):not([aria-disabled="true"])'
+]
+
+unlessNoFocus = (selector) ->
+  "#{selector}#{noFocusSelector.join('')}"
```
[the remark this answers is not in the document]
omargfh: Yeah, no, that's a miss on my end.

app/assets/javascripts/beak/widgets/accessibility/utils.coffee
omargfh: Not a big fan of `Object.fromEntries(Object.entries(object).map(...))`

app/assets/javascripts/beak/widgets/event-traffic-control.coffee
omargfh: it is a dev error. You are intended to catch this before merging into `main`. Users should not be alerted.

app/assets/javascripts/beak/widgets/accessibility/key-combo.coffee
omargfh: Object keys in JavaScript need not be only strings: since ES6, keys are of type `string | Symbol`. Further, object notation coalesces anything that is not a string to a string using its `prototype.toString` method. A common pattern is custom classes that evaluate keys internally using `toString,` and its useful to specify those as well. One can say that keys are of type `string | Symbol | object` noting that objects are not keyed on object identity (but they could be*).

```js
function id() {
  return Math.random().toString(36).slice(2); // simple unique ID
}

class ObjectWithIdentity {
  constructor(...args) {
    Object.assign(this, ...args); // if you want to merge props
    this.id = id();
  }

  toString() {
    return String(this.id);
  }
}

const x = new ObjectWithIdentity({ ... })
const y = new ObjectWithIdentity({ ... })

const z = { [x]: ..., [y]: ...} // <= This is valid
```

Regarding semantic, TypeScript types these as either `Record<K, V>` where K is the key and V is the value type using the generic pattern or 
```ts
{
  [k: K]: V
}
```
using the indexing pattern. I'll use one of these because `Object[String]` is ambigious.

app/assets/javascripts/beak/widgets/accessibility/key-combo.coffee
omargfh: They were originally getters... well until I discovered there is no explicit getter syntax in CoffeeScript.

app/assets/javascripts/beak/widgets/accessibility/keybind.coffee
omargfh: `ractive` is to be treated as a generic of the type `Ractive`. It must be the case that you use the same `Ractive` in all usages of the class. I think the same applies to `Mousetrap` now that I think about it.

omargfh: Changed to 
```coffee
# Keybind[mousetrap, ractive]
# mousetrap: Mousetrap
# ractive: Ractive
export class Keybind
```

app/assets/javascripts/beak/widgets/ractives/tab.coffee
omargfh: What's wrong with `if/else`?

app/assets/javascripts/beak/widgets/ractives/widget.coffee
```diff
@@ -167,10 +178,33 @@ RactiveWidget = RactiveDraggableAndContextable.extend({
...
+        ), 0)
+      return
+
     init: ->
       @findComponent('editForm')?.fire("activate-cloaking-device")
       return
 
+    'on-editor-overlay-keydown': (event, trueEvent) ->
+      if trueEvent.target isnt @find('.editor-overlay')
+        return
```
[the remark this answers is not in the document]
omargfh: Yes but the predicates are long. Prefer them separated. Otherwise it'd be
`if trueEvent.target is @find('.editor-overlay') and  trueEvent.key is 'Enter' and @get('isSelected')`

LaCuneta:
> 👍 `Alt-t` seems to work, `Shift-Alt-t` does nothing for me on Chrome/Firefox+Ubuntu.

I have thought about this a bit more, and with the shortcuts to jump to command center, code, and info, I'm not sure tabbing through those in order even makes much sense, but it's fine to have the option.  What I do really want when I'm "done" with the code tab or command center is usually to jump back to the widgets/interface.  So I think a `Ctrl-0` to jump focus to the first widget might make sense to have.

> The mock tab `Alt-t` and `Shift-Alt-t` should probably also be moved to the code editor list in the new popup along with find usages and comment/uncomment.

I see that this is also used to break out of the command center input where `Tab` changes the context the code will run it.  I think it still might be better to have the hotkey listed in both places, then.

app/assets/javascripts/beak/widgets/event-traffic-control.coffee
TheBizzle: I'm struggling to imagine a scenario where a bug exists and developers need to be made aware of it, but we wouldn't want users to know inform us if they encountered it.

app/assets/javascripts/beak/widgets/event-traffic-control.coffee
```diff
@@ -233,27 +241,63 @@ controlEventTraffic = (controller, performUpdate) ->
...
+      return
+
+    showPropertyName = "show#{tabCanonicalName}"
+
+    if options.active is 'toggle'
+      showPropertyValue = not ractive.get(showPropertyName)
+    else if typeof options.active isnt 'undefined'
+      showPropertyValue = options.active
+    else
+      return
```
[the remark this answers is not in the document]
TheBizzle: I spoke elsewhere in my review about how I object to early exits.

app/assets/javascripts/beak/widgets/accessibility/key-combo.coffee
TheBizzle: You sure?  It didn't when I tested it in the REPL.

app/assets/javascripts/beak/widgets/ractives/tab.coffee
TheBizzle: I agree entirely.  `if`/`else` should be used, instead of simply `if`.

That is, this code should be:

```coffee
if event.key is [snip]
  [snip of side-effects]
  false
else
  true
```

app/assets/javascripts/beak/widgets/ractives/widget.coffee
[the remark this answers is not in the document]
TheBizzle: You can deal with this by storing them in variables with short names.

omargfh:
@LaCuneta 
> - When I tab to the `go` button of a model and hit space bar or enter, nothing happens. On Firefox, space bar does a page down. I assume this is something particular to forever buttons, since the `setup` buttons work fine.

Fixed in caebfa6.

> When I click on a button widget with the mouse, it doesn't get the focus highlight but the button is set as the current element for keyboard access. As such I don't know what will happen if I start trying to tab around. I think it should get the visual focus, too.

HTML differentiates `:focus` and `:focus-visible` psuedo-selectors. Ideally, you either want to tab or use a mouse, not both. I would not like highlighting everything on focus (try it with any web app).

> When I Tab through a slider, it lets me adjust the actual slider with left-right arrows. But I cannot tab to the number entry to type in a number. It feels like I should be able to do either, tab through both slider controls.

When we updated the styling of widgets, we decided that the number entry would not be focusable to allow for left-right and up-down arrows to work by default. I added `Meta+I` to switch between the slider and number entry.

> "You should be able to escape text areas (e.g. the Code tab's editor) to resume tabbing through the page" - this is marked as complete in the issue, but doesn't work for me. Hitting Esc doesn't do anything in the model code editor.

I'm going through your entire comment, so for the sake of completeness: alt+T and shift+alt+T are mock keybinds for Tab and Shift+Tab to move focus out of the code editor. Esc is not a keybind for this action as it may be captured by the code editor itself. I did a patch for a related bug in 5c6d6870.

> If I open the code tab, hit Ctrl-1 to go to the command center, things work as expected, it opens and I can type a command. If I hit Ctrl-2 to go back to the still-open code tab, the code tab closes instead.

Discussed.

> If I have the command center open and then start trying to tab to the model widgets, focus can bounce over to the Clear button in the command center. With Wave Machine it went from the setup button to Clear. This didn't occur every time, but it definitely did happen, most common was after playing with the Ctrl-# hotkeys for a bit.

[CANNOT REPRODUCE] [EXPECTED] I can't reproduce this. Also, from the sound of it, this is the expected behavior. Clear comes after the command center input in the tab order.

> The authoring/interactive Mode button and Commands and Code placement button are not in the tab order, I never get the chance to access them via keyboard shortcut. The Mode actually has a hotkey combination, but it still seems odd you can't get there by tabbing.

Added new global on-activateClick event to toggle these buttons.

> Input widgets with type of String (reporter) and String (command) allow tabbing for focus, but you can't actually edit them; you still have to click to the edit window before can type.

Fixed in 28fe74a0. In particular, the code editor now receives the correct tabindex so it can be focused.

> Related: If I tab through all the widgets in the model, it then goes through many of the top buttons (powered by NetLogo, File: New, etc), but after getting through those (speed slider last), it then goes to the actual code editor control of the String (reporter) input widget! That code editor control then starts capturing tabs to turn them into, well, textual tabs, so I cannot get out of that input once I'm there. All Widgets model in demomodels is a good one to test with.

Fixed in 28fe74a0. The code editor used to receive to set internal tabindex of 0 at all times. In addition, Esc->Tab and Alt+Tab can be both used to exit the code editor.

> When I load a model using Launch and it's ready, I first hit tab and it goes to setup (or the top-leftist widget) on most models I've tried. This seems totally fine. If I use Shift-Tab to go backwards from here the focus goes up to Upload a Model and then the models library drop-down, then up to the top nav-bar manu. I would've expected it to go through the model speel slider, mode button, commands and code setting, and model export/file buttons first before getting up to those other choices. Note that if I tab "down" through the widgets all the way, once I get through the last widget it does circle back to the top of the model, the powered by NetLogo link, then the model file options, etc.

[TODO]

> If I Shift-Tab to get up to the top nav-bar menu, Documentation isn't selectable, but the rest are. I assume this is because the docs entry pops out a dropdown, but it should still be keyboard accessible, I'd think.

Fixed.

> Output Widgets and Monitors are tabbable, which is fine, but you can't actually do anything when they are selected. Maybe letting Ctrl-C copy their current value when selected would be useful to users?

Added.

> We should add the new hotkeys for command center, code, and info to the quick help popup displayed when you press ?

Discussed.

> We should probably add a link/button for the keyboard quick-help that also displays it's keyboard shortcut. To me it's not very discoverable.

[TODO]

> Nice! I'd definitely consider reversing those functions (Ctrl-1 to focus, Ctrl-Shift-1 to open/close), though. I think wanting to focus will be much more common than wanting to close.

Done.

> Ah, neat! The ? key does still show the old popup, though. That should either be removed or set to show the same info as F1. Esc does not banish the popup as it did for the prior one. Ctrl-H (which I see in the new listing) also seems to do the same thing as ?, which seems odd. And on Chrome+Ubunut (not Firefox) hitting F1 does bring the new help window up, but it also brings up Chrome's help window in a new tab. The mock tab Alt-t and Shift-Alt-t should probably also be moved to the code editor list in the new popup along with find usages and comment/uncomment.

[TODO]

> I have thought about this a bit more, and with the shortcuts to jump to command center, code, and info, I'm not sure tabbing through those in order even makes much sense, but it's fine to have the option. What I do really want when I'm "done" with the code tab or command center is usually to jump back to the widgets/interface. So I think a Ctrl-0 to jump focus to the first widget might make sense to have.

Done.

> I see that this is also used to break out of the command center input where Tab changes the context the code will run it. I think it still might be better to have the hotkey listed in both places, then.

[TODO]

omargfh:
Navigating between code tabs and the main widget area is still a bit unusual. Same goes for alerts and anything outside the main widget area. It is a lot more usable now though. I am pretty satisfied with the widget area. I added some shortcuts that makes it usable from the keyboard. In particular, `mod+alt+a` opens the add widget menu, context menu is focusable and focuses by default, edit form focuses first input by default, fixed issues with focus visibility, and added a `shift+{up,down,left,right}` for faster nudging.

https://streamable.com/nrqoe5

app/assets/javascripts/beak/widgets/event-traffic-control.coffee
TheBizzle: This should be made consistent.

app/assets/javascripts/beak/widgets/ractives/tab.coffee
```diff
@@ -0,0 +1,78 @@
+
+
+RactiveTabWidget = Ractive.extend({
+
+  data: () -> {
+      name:             undefined  # String
+    , title:            undefined  # String
+    , show:             false      # Boolean
+    , 'focus-target':   undefined  # String (Selector)
+    # https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView
```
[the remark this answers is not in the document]
omargfh: How would you know where it is documented? It could be coming from a library, custom code, or the DOM.

LaCuneta (changes requested):
Nice work moving the authoring bits forward, that's very good to have!  Thank you for your work on this.

It looks like there are now some conflicts on rebasing - let me know if you have questions on what I did on `main`.  I also see a few unresolved comments/discussions.

Here is my testing feedback, in rough order of what I feel is most important to least important.

- The "fuzzy outline" highlight for focus on widgets is hard to see, especially for buttons.  I don't rememeber it being quite that way before.  I have trouble tracking the focus as I tab through widgets, partly because the tab order surprises me a bit (not much to do about that), but also due to my aging eyeballs.  I'd prefer something more bold so it's obvious where focus is on all widgets.
- I think "`?` for keyboard help" might make more sense than "`?` for help" as it's not really giving general help with the application.
- When I load a model, this first thing to get focus when I hit the tab key is **Powered by NetLogo** and then through the model controls, through the command/code/info, then to my browsers controls, the menu bar, and the models library and upload controls before finally I get to the widgets.  I think the widgets should be the first thing in the tab order on load.  The `Ctrl-0` hotkey makes this a lesser issue, but you do have to know about that to use it.
- The `Ctrl-I` shortcut helps, but I still find it really odd I cannot tab directly to a slider and type in a number using the keyboard.  I wouldn't hold this PR up for that, but maybe we should revisit this at a devel meeting?
- I like the copy value for more widgets, it's very handy!  Is the `alert()` display for that intentional, or was it just for debugging?  It's easy enough to dismiss, just a little weird.  Something more tool-tip-like and ephemeral might be better, but again not necessary to merge.
- Copying the widget value doesn't work for output widgets or color input widgets.

LaCuneta (changes requested):
Testing went well.  I'm very excited to be able to use the keyboard when doing development and testing in the future!  It's going to speed things up greatly.  Just a few items here.

- In the CSS files the spaces have been swapped over to tabs.  We prefer to stick to 2 spaces for indentation.
- The **Documentation** menu link does open when I hit enter on it and it stays open, but I cannot select the links in that menu.  The up and down arrows don't do anything once it's open and I can't think of another way to select the links in the menu there (tested Firefox and Chrome).
- The toast popup for copy notifications is hard to see at the bottom-right.  If I have code and commands at the bottom with code or info open I cannot see it at all.  Maybe the top-right would be a better spot for it?  It also throws an `console.log()` message to the dev console from the `copy-current-value` event in `widget.coffee`.

I think once you've got those sorted, you can change this from a draft PR to "ready" and we can look at merging.

TheBizzle (approved):
Quality! 🏆

LaCuneta (approved):
:+1:  Excellent, thank you!

PR #455, opened 2025-11-18
[Fixes #454] Bundle OpenSans font for NetLogo Web

### LINKED ISSUE #454: Missing font in exported HTML page
When I load the exported HTML page of a model, the font is different, and the console says, "Failed to load resource: OpenSans-VariableFont...". Is there any way that NetLogo Web could convert that font reference to a Google Fonts link on export, or something like that?

omargfh:
This PR fixes issue #454.

This PR is compliant with license requirement listed below:

```txt
...
Original or Modified Versions of the Font Software may be bundled,
redistributed and/or sold with any software, provided that each copy
contains the above copyright notice and this license. These can be
included either as stand-alone text files, human-readable headers or
in the appropriate machine-readable metadata fields within text or
binary files as long as those fields can be easily viewed by the user.
...
```

Compliance is met via 
1. Comment prepended to HTML exports with license content (HTML Export) 
>  human-readable headers or in the appropriate machine-readable metadata fields
within text or binary files as long as those fields can be easily viewed by the user
2. URL license linking on the `netlogoweb.org` website 
>  can be included either as stand-alone text files

:warning: Please make the following asset URL publicly accessible: https://www.netlogoweb.org/fonts/Open_Sans/OFL.txt


LaCuneta (changes requested):
Attributions setup looks good to me.  I tested it out and bundling worked as expected.  

- Can you add a comment to the CSS just noting which file was the source for the base64 string and why it's being embedded like this?
- Having the font embedded in the `netlogoweb.css` makes it harder to inspect and edit that file; can we move the embedded fonts into their own CSS file?

Thanks!

LaCuneta (approved):

Issue #463, opened 2026-09-15
No contrast in anchor links inside a toast


omargfh:
# No contrast for anchor links inside a toast

Discovered while addressing [#461](https://github.com/NetLogo/Galapagos/issues/461).

##  Steps to recreate

1. Navigate to [NetLogo Web Launch](https://netlogoweb.org/launch)
2. Load a model, make a change, and reload the page
3. Observe the "Load Changes" toast
4. Expect the anchor link inside the toast to be unreadable against the toast background

## Investigation

Toasts do not style anchors, so links inside them inherit the page's default link color, which has no contrast against the toast background. Any toast containing a link is affected, not just the "Load Changes" toast.

This bug will be addressed in the same PR as [#461](https://github.com/NetLogo/Galapagos/issues/461).

Issue #462, opened 2026-09-15
Engine crashes when loading a WIP `.nlogo` file


omargfh:
# Engine crashes when loading a WIP `.nlogo` file

Discovered while addressing #461.

## Steps to recreate
1. Navigate to [NetLogo Web Launch](https://netlogoweb.org/launch)
2. Load a `.nlogo` model file 
3. Enter authoring mode and make some changes
4. Reload the page and load the same `.nlogo` file again
5. Expect the engine to crash

## Error Message
> There was an error compiling the model's code. The given model was not a NetLogo 7 .nlogox file or a NetLogo 6.4 .nlogo file.
>
> Check that you selected the correct file to upload.
>
> The model contents that failed to compile are given below.
>
> ```xml
> <?xml version="1.0" encoding="UTF-8"?>
> <model version="NetLogo 7.0.4" nlw-version="2.16.0 (4690bbb3)">
>   ...
> </model>
> ```

## Investigation
The crash is a format mismatch between the original file and the cached changes. 

1.  When a model is edited, the _work-in-progress listener_ stores the result of `SessionLite.getNlogo()`, which always calls `compiler.exportNlogoXML`. So the cached copy is nlogox XML regardless of the format the model was loaded from.
2.  On the next load, `Tortoise.fromNlogo` decides which parser to use by calling `nlogoSource.isOldFormat()` on the original source. For a legacy `.nlogo` file that returns `true`, so the cached XML is handed to `fromNlogoSync`.

This bug will be addressed in the same PR as #461.

PR #464, opened 2026-09-15
[Fixes #461, #462, #463] Make loading auto-saved model changes opt-in

### LINKED ISSUE #461: Make auto-saved model loading opt-in rather than automatic.
As discussed in the development meeting conducted on September 9, 2026:

- [ ] Disable loading auto-saved models by default
- [ ] Replace the existing "revert to original" toast with "load changes" toast
- [ ] Remove the settings option "Automatically save changes made to models and reuse them when reloading" as it becomes irrelevant 
- [ ] Add a new settings option to control the "load changes" toast behavior to give users an opt-out for spam

omargfh:
# Make loading auto-saved model changes opt-in

Fixes #461, #462, and #463.

## Changes

**Behavior**

- Cached changes are no longer applied automatically on load. When a cached copy exists, a "Load Changes" toast offers to apply it. The "revert to original" toast is removed.
- Removed the setting "Automatically save changes made to models and reuse them when reloading." Changes are always saved.
- Added a setting to hide the "Load Changes" toast. Defaults to shown.

**API**

- `fromNlogo` takes `workInProgressNlogo: String | null` and compiles whatever it is handed. `newSession` no longer takes `workInProgressState`.
- `WorkInProgressState` gains `enabled-with-unloaded-wip`.
- `loadWip` returns `WipInfo | null`.
- `_syncState` returns the state so `setSession` reads it once.

**Call sites**

- `loadWorkInProgress` in `simulation.js` guards against a null listener or an empty cache.
- `nlw-revert-wip` collapses to `loadSource(wipListener.getNlogoSource(), null)`; `revertWip` clears the source title.
- HubNet Web's `loadHubNetWeb` updated for the new `newSession` signature.

## Why WIP handling moved out of Tortoise

Tortoise used to own WIP state because cached changes were applied at load time, so the loader was the only place that knew both which nlogo it compiled and whether it came from the cache. It derived `workInProgressState` and passed it through `newSession` so the Revert button rendered correctly from the first frame, and the listener sometimes had to patch that state afterwards.

Opt-in loading separates the two questions. The loader now compiles whatever `workInProgressNlogo` it is handed, the page decides when to hand it over via `loadWip` and `loadSource`, and `WipListener.getState()` derives state on demand from what only the listener knows. The loader could not compute `enabled-with-unloaded-wip` anyway, since that state never touches the compile path.

## #462 fix

Cached changes are always exported as nlogox XML by `compiler.exportNlogoXML`, but `Tortoise.fromNlogo` was picking the parser based on the format of the original file. Loading a legacy `.nlogo` file with cached changes therefore handed XML to the `.nlogo` parser and crashed the engine.

`fromNlogo` now dispatches on the string it actually compiles, using the new exported `isOldFormatNlogo` helper in `nlogo-source.coffee`.

## #463 fix

Anchor links inside toasts inherited the page's default link color and were unreadable against the toast background. Toast anchors now get their own color and underline so they read at any toast variant.

## Side effects

- NetTango `fromURL` now passes `null` for WIP, so URL-loaded NetTango models show `disabled` instead of `enabled` with an empty cache. This hides the Revert button, which was always disabled in that case anyway.

## Known limitations

- The old `workInProgress.enabled = false` setting is ignored. No storage version bump. Why? The current storage version is `1`, which means no one had tried to update it before. I recall Jeremy's recent work on pinning engine versions, and I can imagine a scenario where the version stored is higher than the one expected. 

## Testing

Chrome and Firefox:

- [x] Open a model, move a widget, refresh, click Load Changes. Changes load.
- [x] Open a model, move a widget, refresh, edit the model again, refresh, click Load Changes. Latest changes load.
- [x] Open a model, move a widget with the keyboard, refresh, click Load Changes.
- [x] Open a model, move a widget, refresh while in authoring mode.
- [x] Load `Patch and World Size.nlogo`, edit, refresh, click Load Changes. No crash (#462).
- [x] HubNet Web join.
- [x] `demomodels/All Widgets`.
- [x] `coffee` compiles
- [x] `coffeelint` and `test`
- [x] `node --check`


LaCuneta (changes requested):
Thank you for your work on this!  I really like that I can keep using my existing model edits without much friction.  I also appreciate the bug fixes.

Three issues I found:

1. If you open a model with cached changes, then select to load those changes, things work as expected.  But if I then upload a model that also has cached changes, I do not get prompted to apply them.  If I refresh the page and re-upload, it does work, so it seems like the page state is getting set by the first model loaded from cache. 
2. This is a silly edge case, but if you open a model with cached changes then use your fleetest fingers to make an edit while the Load Changes toast is up, when you hit Load Changes it'll load the change you just made, not the one that was in the cache when the toast was shown.  I think my preference is to just hide the toast early if someone dives into editing right away.
3. Previously we displayed the source of the model (switch code from `createWorkInProgressAlert()` when talking about it in the toast, now all models are "dismiss to keep the model as linked".  I think being more specific is helpful, but I don't consider it a blocker to merge.

omargfh:
Hi @LaCuneta, thank you for taking the time to review this PR. I am glad it is going smooth.

The devel direction right now is to get rid of the toast in favor of highlighting the `Load Changes` button.

I'll look into the remaining concerns and address them shortly.

omargfh:
> 1. If you open a model with cached changes, then select to load those changes, things work as expected.  But if I then upload a model that also has cached changes, I do not get prompted to apply them.  If I refresh the page and re-upload, it does work, so it seems like the page state is getting set by the first model loaded from cache.

Toast no longer exists.

> 2. This is a silly edge case, but if you open a model with cached changes then use your fleetest fingers to make an edit while the Load Changes toast is up, when you hit Load Changes it'll load the change you just made, not the one that was in the cache when the toast was shown.  I think my preference is to just hide the toast early if someone dives into editing right away.

Toast no longer exists.

> 3. Previously we displayed the source of the model (switch code from `createWorkInProgressAlert()` when talking about it in the toast, now all models are "dismiss to keep the model as linked".  I think being more specific is helpful, but I don't consider it a blocker to merge.

Toast no longer exists but I added it as a hover title on the "Load Changes" button.

LaCuneta (approved):
Nice, looks good to me.  The "load model, Load Changes, upload a model, don't see cached changes" bug is still there even without the toast, but I'll go ahead and merge this as I noticed a related existing bug with reverting that'll need fixing anyways. I'll take care of the other at the same time.

